PR #1, opened 2020-02-28
adding experiment for docker-compose setup and related scripts

grexe:
as promised, here are my docker-compose scripts for devs who do not want to go all the way to GCE

### DIFF
--- a/chaos-experiments/docker-compose/broker-outage/experiment.json
+++ b/chaos-experiments/docker-compose/broker-outage/experiment.json
@@ -16,8 +16,11 @@
         "tolerance": 0,
         "provider": {
           "type": "process",
-          "path": "../../scripts/docker-compose/verify-steady-state.sh",
+          "path": "../scripts/verify-steady-state.sh",
           "timeout": 60
+        },
+        "pauses": {
+          "after": 5
         }
       }]
   },
@@ -27,8 +30,21 @@
       "name": "terminate-random-broker",
       "provider": {
         "type": "process",
-        "path": "../../scripts/docker-compose/kill-random-broker.sh"
+        "path": "../scripts/kill-random-broker.sh"
+      },
+      "pauses": {
+        "after": 5
       }
-    }],
-  "rollbacks": []
+    }
+  ],
+  "rollbacks": [
+    {
+      "type": "action",
+      "name": "restart-broker",
+      "provider": {
+        "type": "process",
+        "path": "../scripts/start-broker.sh"
+      }
+    }
+  ]
 }
--- a/chaos-experiments/docker-compose/broker-outage/experiment.json
+++ b/chaos-experiments/docker-compose/broker-outage/experiment.json
@@ -1,0 +1,34 @@
+{
+  "version": "0.1.0",
+  "title": "Zeebe broker outage experiment",
+  "description": "Zeebe should be fault-tolerant when one broker dies",
+  "contributions": {
+    "reliability": "high",
+    "availability": "high"
+  },
+  "steady-state-hypothesis":
+  {
+    "title": "Zeebe is alive",
+    "probes": [
+      {
+        "name": "should-be-able-to-create-workflow-instances",
+        "type": "probe",
+        "tolerance": 0,
+        "provider": {
+          "type": "process",
+          "path": "../../scripts/docker-compose/verify-steady-state.sh",
+          "timeout": 60
+        }
+      }]
+  },
+  "method": [
+    {
+      "type": "action",
+      "name": "terminate-random-broker",
+      "provider": {
+        "type": "process",
+        "path": "../../scripts/docker-compose/kill-random-broker.sh"
+      }
+    }],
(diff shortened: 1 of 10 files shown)


Zelldon:
Hey @grexe 

this is the first PR yeah :tada: Thank you very much.

Actually we want to run all of our chaos experiments on kubernetes, since this our supported environment and nearer to the actual production setup's. Running a cluster on a local machine is also really resource limited. Another reason for using kubernetes is that we want to automate all of the experiments in our gke cluster. How a benchmark setup normally looks like will be added in the normal Zeebe Repository, please see the related PR https://github.com/zeebe-io/zeebe/pull/3953 . 

That is the reason why I'm not 100% sure if this makes sense to add here, but the experiments we have can be an inspiration for others to create there own tests.

Does it make sense to you?

Greets
Chris

grexe:
Hehe also glad to issue the first PR here:)

Well I think we have 2 different use cases here:
1. developers at Zeebe/camunda who use a common environment for internal testing and
1. developers outside of Camunda/Zeebe, who want to perform chaos tests on *their* environment for checking the setup and resiliency.

For 2., a docker-compose setup would be very useful imho.

Also, the benchmarking guide you linked above tackles performance tests using kubernetes again, but not chaos testing.
We are not on Kubernetes yet, still AWS and on-prem only. I suspect other projects will be in a similar situation and not evereybody is on k8s yet...

chaos-experiments/scripts/docker-compose/verify-steady-state.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+cd $HOME/Develop/zeebe/zeebe-performance-tests/src/
```
Zelldon: Could we have something, which really can be executed without depending on external resources? So like using Zbctl. We could add a simple deployment to the repository, which we just deploy and start via zbctl. I think I can add some process, which we could use also in the kubernetes tests then. This would also make sense for the existing tests, such that we are not depending on the benchmark setup.

Furthermore it makes sense to start multiple instances so we make sure we can start instances on all partitions.

chaos-experiments/docker-compose/broker-outage/experiment.json
```diff
@@ -0,0 +1,34 @@
...
+  "method": [
+    {
+      "type": "action",
+      "name": "terminate-random-broker",
+      "provider": {
+        "type": "process",
+        "path": "../../scripts/docker-compose/kill-random-broker.sh"
+      }
+    }],
+  "rollbacks": []
```
Zelldon: Since we expecting docker-compose, I assume this setup https://github.com/zeebe-io/zeebe-docker-compose/blob/master/cluster/docker-compose.yml ? We should bring the container back on rollback. In kubernetes it happens automatically if you delete a pod, which is the reason we don't need it there.

chaos-experiments/scripts/docker-compose/verify-steady-state.sh
grexe: ah yes sure, I was using the zeebe-performance-tests project by Zeebe but shouldn't have added a local dependency, we can probably just use the same test workflow that is used in the k8s setup.

chaos-experiments/docker-compose/broker-outage/experiment.json
grexe: Sure, let me check what docker-compose supports here or if I have to do it manually.

grexe: Ok I've updated my PR and added a clean, self-contained, working chaostoolkit-test setup with docker-compose, rearranged files a bit.

grexe:
Ok I've updated my PR and added a clean, self-contained, working chaostoolkit-test setup with docker-compose, rearranged files a bit.
Please have another look @Zelldon

chaos-experiments/docker-compose/scripts/verify-steady-state.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+echo "deploying test workflow…"
+zbctl --insecure deploy ../../bpmn/diagram_1.bpmn
+
+echo "Starting test workflow run for 60s…"
+zbctl --insecure create instance benchmark
```
Zelldon: Since you have a message catch event, we could try the following out.

Deploy the workflow, send a message, start an instance and await the result. With this we could make sure that the complete cycle still works. What do you think? To await the result we just need to add the `--withResult` flag.

grexe:
yes you're right @Zelldon , totally forgot that (and I could have sworn I tested that;) - added now and runs locally.

Zelldon:
Thanks @grexe for your contribution. :tada: 

I will probably change some minor stuff, like in the `runChaos.sh` using `docker exec zeebe_broker_1 zbctl --insecure status` instead of relying on the users zbctl installation, which might not exist and add a Readme. But this is a good start for now :+1:

PR #47, opened 2021-03-09
chore(scripts): cast workflow instance key to number

pihme:
Fix for: https://github.com/zeebe-io/zeebe-cluster-testbench/issues/247


Zelldon (approved):
Thanks @pihme :+1:

Issue #49, opened 2021-03-23
Deployment distribution experiment is not executed on camunda cloud


Zelldon:
Currently we can't run chaos experiments with network partitions or at least they will not have an effect, since we missing the NET_ADMIN capability, see https://github.com/zeebe-io/zeebe-cluster-testbench/issues/237

@hisImminence was so kind to show me a different way to make the experiments work. Instead of having the capability in general in the controller we will just patch the cluster we want to test.

> Disable reconcile `kubectl patch zb CLUSTERID --type merge --patch='{"spec":{"controller":{"reconcileDisabled":true}}}'`
> Apply patch `kubectl patch statefulset zeebe --patch "$(cat net_admin-patch.yaml)"`
> net_admin-patch.yaml:
> ```yaml
> spec:
>   template:
>     spec:
>       containers:
>         - name: "zeebe"
>           securityContext:
>             capabilities:
>               add:
>                 - "NET_ADMIN" 
> ```

I tested this manually and it works :+1: We need to adjust our chaos scripts.

Zelldon:
@deepthidevaki lets do this today

Zelldon:
I had to revert https://github.com/zeebe-io/zeebe-chaos/commit/123ce9e9e9267419ebf22e71c3aed049d002d24f to unblock testbench. We need to apply the changes after the service account is fixed. 

See https://zeebe-io.github.io/zeebe-chaos/chaos_experiment/camunda_cloud/network/2021/03/23/camunda-cloud-network-partition.html

Zelldon:
works now :rocket:

Zelldon:
```
[2021-03-23 14:21:24 DEBUG] [process:53] Running: ['/home/chaos/zeebe-chaos/chaos-experiments/camunda-cloud/../scripts/apply_net_admin.sh']
[2021-03-23 14:21:25 DEBUG] [__init__:142] Data encoding detected as 'ascii' with a confidence of 1.0
[2021-03-23 14:21:25 DEBUG] [activity:180]   => succeeded with '{'status': 0, 'stdout': 'zeebecluster.cloud.camunda.io/cc108db7-768c-45cc-a6c5-098dc28f260c patched\nstatefulset.apps/zeebe patched\n', 'stderr': ''}'
```

PR #50, opened 2021-03-23
Add deployment distribution experiment with network partition

deepthidevaki:
closes https://github.com/zeebe-io/zeebe-chaos/issues/49


Zelldon (approved):
Thanks so much :tada:

PR #55, opened 2021-04-26
fix(script): strip quotes from message key

korthout:
The json returned by `zbctl publish message` contains a key property whose value is a string (not a number). So, it has quotes around it's value, which need to be stripped before the partitionId can be extracted from the key (using a bitshift).

This PR fixes this problem by stripping any non-digit character from the keyString value.


Zelldon (approved):
giphy

PR #65, opened 2021-11-25
Images are now zoomable

multani:
The images with the charts are quite cool but they are quite small, unless they are opened in a dedicated window.

This makes the images clickable, which will zoom the image in fullscreen.

This uses this plugin: https://github.com/flexanalytics/plugin-image-zoom


Zelldon:
Awesome @multani thanks so much for your contribution! 

I normally made the images clickable but forgot on the recent post again, sorry😅🙈 

Thanks :bow:

PR #66, opened 2021-11-25
Fix the documentation build on GitHub Actions

multani:
It tried to access a Git repository (not published on the npm registry) with the git+ssh protocoal, without having a usable SSH key.

This now uses the (unauthenticated) HTTPS protocol.

It also tries to build the project beforehand, which should show the failure earlier.


Zelldon:
Thanks @multani :rocket:

PR #67, opened 2021-11-25
Remove shellcheck GitHub Actions

multani:
It's failing for a long time now, so I guess we can remove it :)


Zelldon:
If fixed alll shellcheck errors instead :) https://github.com/zeebe-io/zeebe-chaos/pull/69

multani:
> If fixed alll shellcheck errors instead :) #69

That also works ;)

PR #114, opened 2022-02-18
fix: verify gateway pods are ready using deployment

deepthidevaki:
We can use the deployment for `kubectl wait`. The check fails if all pods in the deployment are ready.

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/568

### DIFF
--- a/chaos-workers/chaos-experiments/scripts/verify-readiness.sh
+++ b/chaos-workers/chaos-experiments/scripts/verify-readiness.sh
@@ -8,7 +8,13 @@ namespace=$(getNamespace)
 # verify that everything is running
 
 if [ "${CHAOS_SETUP}" == "cloud" ]; then
-  kubectl wait --for=condition=Ready pod -l app.kubernetes.io/app=zeebe --timeout=900s -n "$namespace"
+  # Wait until brokers are ready. The componenet should be broker, but it is wrongly labelled as gateway
+  kubectl wait --for=condition=Ready pod -l app.kubernetes.io/component=gateway --timeout=900s -n "$namespace"
+  # Wait until all gateway pods are running.
+  kubectl wait --for=condition=Available deployment zeebe-gateway --timeout=120s -n "$namespace"
 else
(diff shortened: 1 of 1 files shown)


chaos-workers/chaos-experiments/scripts/verify-readiness.sh
```diff
@@ -8,7 +8,13 @@ namespace=$(getNamespace)
 # verify that everything is running
 
 if [ "${CHAOS_SETUP}" == "cloud" ]; then
-  kubectl wait --for=condition=Ready pod -l app.kubernetes.io/app=zeebe --timeout=900s -n "$namespace"
+  # Wait until brokers are ready. The componenet should be broker, but it is wrongly labelled as gateway
+  kubectl wait --for=condition=Ready pod -l app.kubernetes.io/component=gateway --timeout=900s -n "$namespace"
+  # Wait until all gateway pods are running.
+  kubectl wait --for=condition=Available deployment zeebe-gateway --timeout=120s -n "$namespace"
```
Zelldon, line 14: It would be nice if we could add a why in the comments here. So why we have to use the wait for pod (broker) ready PLUS deployment avail for gateway.

Zelldon (approved):
Thanks @deepthidevaki :+1: 

Tbh I think it can't be the end solution, since we only validate that the deployment is available. But I see no other way currently, with the preemtable nodes and the kubectl wait. Thanks for fixing :bow: 

Please adjust the comment and then we can merge it. Afterwards we need to create a new release :)

deepthidevaki:
> Tbh I think it can't be the end solution, since we only validate that the deployment is available.

Why do you think so? Deployment is available only if all pods are running. If a pod is terminated, then this check fails. According to me this is exactly what we want to check.

Zelldon:
Thanks for me it was not 100% clear, since deployment and pods are decoupled, I thought. So my fear was that the deployment is available/ deployed but not all pods are ready. 

It seems to be related to the minimum replica config https://mrkaran.dev/posts/kubectl-wait/

deepthidevaki:
Oops. I merged before committing the changes :face_with_head_bandage: 
Created a new PR https://github.com/zeebe-io/zeebe-chaos/pull/114

PR #116, opened 2022-02-22
fix: don't try to use shutdown gateways

oleschoenburg:
When gateway pods are evicted, they stay around with status "Shutdown".
We need to filter for running pods to get a gateway that we can actually use.

### DIFF
--- a/chaos-workers/chaos-experiments/scripts/utils.sh
+++ b/chaos-workers/chaos-experiments/scripts/utils.sh
@@ -61,7 +61,7 @@ function getBroker()
 function getGateway()
 {
   namespace=$(getNamespace)
-  pod=$(kubectl get pod -n "$namespace" $(getGatewayLabels) -o jsonpath="{.items[0].metadata.name}")
+  pod=$(kubectl get pod -n "$namespace" --field-selector status.phase=Running $(getGatewayLabels) -o jsonpath="{.items[0].metadata.name}")
 
   echo "$pod"
 }
--- a/chaos-workers/chaos-experiments/scripts/utils.sh
+++ b/chaos-workers/chaos-experiments/scripts/utils.sh
@@ -38,7 +38,7 @@ function runOnAllBrokers()
 {
   namespace=$(getNamespace)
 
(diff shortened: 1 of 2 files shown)


chaos-workers/chaos-experiments/scripts/utils.sh
```diff
@@ -38,7 +38,7 @@ function runOnAllBrokers()
 {
   namespace=$(getNamespace)
 
-  pods=$(kubectl get pod -n "$namespace" $(getBrokerLabels) -o jsonpath="{.items[*].metadata.name}")
+  pods=$(kubectl get pod -n "$namespace" "$(getBrokerLabels)" -o jsonpath="{.items[*].metadata.name}")
```
Zelldon, line 41: :x: this quoating doesn't work here. We want to have multiple labels here, splitted. Either we ignore this shellcheck error or we implement this differently.

chaos-workers/chaos-experiments/scripts/utils.sh
```diff
@@ -53,15 +53,15 @@ function getBroker()
...
+  pod=$(kubectl get pod -n "$namespace" "$(getBrokerLabels)" -o jsonpath="{.items[$index].metadata.name}")
 
   echo "$pod"
 }
 
 function getGateway()
 {
   namespace=$(getNamespace)
-  pod=$(kubectl get pod -n "$namespace" $(getGatewayLabels) -o jsonpath="{.items[0].metadata.name}")
+  pod=$(kubectl get pod -n "$namespace" --field-selector status.phase=Running "$(getGatewayLabels)" -o jsonpath="{.items[0].metadata.name}")
```
Zelldon, line 64: Cool :+1:

Zelldon (changes requested):
Thanks @oleschoenburg :) Please check the quoating stuff again. You could also try it locally. If your source the `util.sh` (`source utils.sh`) can call the functions you will see what I mean.

chaos-workers/chaos-experiments/scripts/utils.sh
oleschoenburg: Oh, of course :facepalm: I'll add an ignore instead.

Zelldon (approved):
:medal_military:

PR #166, opened 2022-08-26
Publish message on specific partition

Zelldon:
Adds a new feature to publish a message with a printable correlation key to an specific (given) partition. This is useful if we want to experiment with message correlation, for example disconnect certain partitions etc.

```go
// The message publish partition distribution is based
// on the following hashcode and modulo by the partition count
//
// Since this calculation is used not only on publish, but also for
// opening the message and instance subscriptions we must support for
// backwards compatibility, otherwise message will not be correlated
//
// This allows us to make certain assumptions, and shortcuts.
//
// static int getSubscriptionHashCode(final DirectBuffer correlationKey) {
//   is equal to java.lang.String#hashCode
//   int hashCode = 0;
//
//   for (int i = 0, length = correlationKey.capacity(); i < length; i++) {
//      	hashCode = 31 * hashCode + correlationKey.getByte(i);
//   }
//   return hashCode;
// }
//
// and
//   public static int getSubscriptionPartitionId(
//      final DirectBuffer correlationKey, final int partitionCount) {
//    final int hashCode = getSubscriptionHashCode(correlationKey);
//    // partition ids range from START_PARTITION_ID .. START_PARTITION_ID + partitionCount
//    return Math.abs(hashCode % partitionCount) + START_PARTITION_ID;
//  }
// Based on the hash code function, we now that if the string has a length of one byte
// no multiplication will happen.
//
// Since we need a printable (for variables) character we start with ASCII code 48, until 126.
// Formulas:
//
// expectedPartition = (hashcode mod partitionCount) + partitionStartId
// hashcode = i ( 48 >= i >= 126) == ASCII CODE
//
```

# Usage

```sh
$ ./zbchaos publish -v --partitionId 1
Connecting to zell-chaos
Send message 'msg', with correaltion key '0' (ASCII: 48) 
Message was sent and returned key 2251799813685346, which corresponds to partition: 1
```

### DIFF
--- a/go-chaos/cmd/publish.go
+++ b/go-chaos/cmd/publish.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/zeebe-io/zeebe-chaos/go-chaos/internal"
@@ -57,7 +58,7 @@ var publishCmd = &cobra.Command{
 			fmt.Printf("Send message with correaltion key '%s' (ASCII: %d) \n", correlationKey, int(correlationKey[0]))
 		}
 
-		messageResponse, err := zbClient.NewPublishMessageCommand().MessageName(msgName).CorrelationKey(correlationKey).Send(context.TODO())
+		messageResponse, err := zbClient.NewPublishMessageCommand().MessageName(msgName).CorrelationKey(correlationKey).TimeToLive(time.Minute * 5).Send(context.TODO())
 		partitionIdFromKey := internal.ExtractPartitionIdFromKey(messageResponse.Key)
 
 		if Verbose {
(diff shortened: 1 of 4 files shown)


go-chaos/cmd/publish.go
```diff
@@ -0,0 +1,72 @@
...
+
+		if Verbose {
+			fmt.Printf("Send message '%s', with correaltion key '%s' (ASCII: %d) \n", msgName, correlationKey, int(correlationKey[0]))
+		}
+
+		messageResponse, err := zbClient.NewPublishMessageCommand().MessageName(msgName).CorrelationKey(correlationKey).TimeToLive(time.Minute * 5).Send(context.TODO())
+		partitionIdFromKey := internal.ExtractPartitionIdFromKey(messageResponse.Key)
+
+		if Verbose {
+			fmt.Printf("Message was sent and returned key %d, which corresponds to partition: %d\n", messageResponse.Key, partitionIdFromKey)
```
npepinpe, line 63: :+1: Very nice. I'm wondering if it shouldn't always be printed regardless, as if a test were to fail due to a partition mismatch, it would be pretty important info

npepinpe (approved):
:+1: 

Just one comment about verbose logging

PR #167, opened 2022-08-30
Allow to specify other process models

Zelldon:
This feature allows specifying different process models for the `verify steady-state` command. This is useful for other experiments, for example where I want to deploy a model with a message catch event.

### DIFF
--- a/go-chaos/internal/zeebe.go
+++ b/go-chaos/internal/zeebe.go
@@ -19,6 +19,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/camunda-cloud/zeebe/clients/go/pkg/pb"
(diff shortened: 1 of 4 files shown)


go-chaos/internal/zeebe.go
```diff
@@ -130,25 +131,45 @@ func extractNodeId(topologyResponse *pb.TopologyResponse, partitionId int, role
...
 	return nil
 }
 
+// if file not exist we read our default BPMN process model and return the content
+func readBPMNFileOrDefault(fileName string) ([]byte, error) {
+	var bpmnBytes []byte
+	var err error
+
+	if len(fileName) == 0 {
+		fileName = "bpmn/one_task.bpmn"
```
npepinpe, line 157: :wrench: I would add verbose logging that we're falling back to the default process since nothing was given. It might seem redundant to you here because we also log later which process we use, but if someone doesn't know the internals of zeebe-chaos then they wouldn't know why this was picked.

npepinpe (approved):
:+1:

Zelldon:
supersed by https://github.com/zeebe-io/zeebe-chaos/pull/174

PR #174, opened 2022-09-21
Make message chaos experiment possible

Zelldon:
Before my surgery I run a new message correlation experiment, details you can find in the blog https://zeebe-io.github.io/zeebe-chaos/2022/08/31/Message-Correlation-after-Network-Partition For that experiment I had to add some new features to the zbchaos toolkit

I already opened some PR's #166 #167 but this was only part of it, so now here is everything I added to make this experiment possible.

* Allow specifying different process models for the verify steady-state command. This is useful for other experiments, for example where I want to deploy a model with a message catch event.
* Allow to specify the message name
* Allow to specify variables on the process instance creation
* Iterate over the message publish correlation key, we use now printable characters otherwise they cant be used in variables

### DIFF
--- a/go-chaos/internal/zeebe.go
+++ b/go-chaos/internal/zeebe.go
@@ -136,9 +136,13 @@ func extractNodeId(topologyResponse *pb.TopologyResponse, partitionId int, role 
 var bpmnContent embed.FS
 
 func DeployModel(client zbc.Client, fileName string) error {
-	bpmnBytes, err := readBPMNFileOrDefault(fileName)
+	bpmnBytes, fileName, err := readBPMNFileOrDefault(fileName)
 	if err != nil {
 		return err
+	}
+
+	if Verbosity {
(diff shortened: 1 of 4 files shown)


go-chaos/internal/zeebe.go
```diff
@@ -134,26 +135,51 @@ func extractNodeId(topologyResponse *pb.TopologyResponse, partitionId int, role
...
+	return processDefinitionKey, nil
+}
+
+// if file not exist we read our default BPMN process model and return the content
+func readBPMNFileOrDefault(fileName string) ([]byte, string, error) {
+	var bpmnBytes []byte
+	var err error
+
+	if len(fileName) == 0 {
+		fileName = "bpmn/one_task.bpmn"
```
Zelldon, line 166: See https://github.com/zeebe-io/zeebe-chaos/pull/167#discussion_r959546145

oleschoenburg (approved):

PR #191, opened 2022-10-12
Go zbchaos worker

oleschoenburg:
This adds a new command `zbchaos worker` that starts a worker that listens to `zbchaos` jobs from testbench and executes them. To execute, we don't spawn a new process and instead use corba to directly call the command with the provider arguments.


go-chaos/go.sum
```diff
@@ -128,7 +128,7 @@ github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44/go.mod h1:bbYlZJ7
 github.com/bugsnag/bugsnag-go v0.0.0-20141110184014-b1d153021fcd/go.mod h1:2oa8nejYd4cQ/b0hMIopN0lCRxU0bueqREvZLWFrtK8=
 github.com/bugsnag/osext v0.0.0-20130617224835-0dd3f918b21b/go.mod h1:obH5gd0BsqsP2LwDJ9aOkm/6J86V6lyAXCoQWGw3K50=
 github.com/bugsnag/panicwrap v0.0.0-20151223152923-e2c28503fcd0/go.mod h1:D/8v3kj0zr8ZAKg1AQ6crr+5VwKN5eIywRkfhyM/+dE=
-github.com/camunda-cloud/zeebe/clients/go v1.3.14 h1:7YR0fAfn2b7X0+QR5SUd4heGb+TPY5kSXy9E3529b6M=
+github.com/camunda-cloud/zeebe/clients/go v1.3.14 h1:NjvinpaQIUlQQbfseqWFrCGkFB8SnbrG3fNongTzac8=
```
oleschoenburg, line 131: :question: I'm a bit confused about this one, without fixing the checksum I can't download the dependencies. Not sure why the checksum would have changed.

oleschoenburg: Yeah, CI says that the issue is on my side. I've fixed it by cleaning up my $GOPATH. I'll remove the commit from here.

oleschoenburg: Actually nevermind, I still have the same issue after cleaning up $GOPATH. I'll upgrade to 8.1 then.

oleschoenburg:
I had a bunch of problems trying to keep go.sum up  to date with go.mod so I upgraded to go 1.19 too which solved the issue for me :shrug:

Zelldon:
I use also g1.19 locally :+1:

oleschoenburg:
Then I'd say it's ready for review :)

go-chaos/cmd/worker.go
```diff
@@ -0,0 +1,112 @@
...
+	clusterId string
+	provider  ChaosProvider
+}
+
+func start_worker(cmd *cobra.Command, args []string) {
+	credsProvider, err := zbc.NewOAuthCredentialsProvider(&zbc.OAuthProviderConfig{
+		ClientID:               os.Getenv("TESTBENCH_CLIENT_ID"),
+		ClientSecret:           os.Getenv("TESTBENCH_CLIENT_SECRET"),
+		Audience:               strings.TrimSuffix(os.Getenv("TESTBENCH_ADDRESS"), ":443"),
+		AuthorizationServerURL: os.Getenv("TESTBENCH_AUTHORIZATION_SERVER_URL"),
```
Zelldon, line 58: :x: Would be great if we could leave testbench out here. Can you check whether the go client already supports env vars for the cloud client? Maybe this is not event necessary.

The testbench env vars can then be injected into the docker image for example

go-chaos/cmd/worker.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func runZbChaosCommand(args []string, ctx context.Context) error {
+	rootCmd.SetArgs(args)
+	_, err := rootCmd.ExecuteContextC(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
Zelldon, line 112: Thats great :rocket:

go-chaos/go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/zeebe-io/zeebe-chaos/go-chaos
 
-go 1.17
+go 1.19
```
Zelldon, line 3: :rocket:

Zelldon (changes requested):
Awesome thanks for taking the initative @oleschoenburg 
Great work and idea with running the worker as own command. I first had in mind to have a separate application or something which executes the worker, but this might be even better :+1: 

Please see my comment regarding the env vars. Lets fix this first, before merging :)

Zelldon (approved):
Super Thanks :+1:

PR #192, opened 2022-10-12
Build zbchaos docker image

oleschoenburg:
Depends on #191 

Builds a docker image `gcr.io/zeebe-io/zbchaos` that runs zbchaos (with the worker command by default).


go-chaos/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# syntax=docker/dockerfile:1.4
+FROM golang:alpine as builder
+
+WORKDIR /app 
+
+COPY --link go.mod go.sum .
+RUN go mod download
+
+COPY --link cmd/ ./cmd
```
Zelldon, line 9: TIL: --link creates a symlink? Wasnt aware of this feature :+1:

Zelldon (approved):
Mega :rocket:

go-chaos/Dockerfile
oleschoenburg: Not exactly but it allows to copy files independently from previous layers: https://docs.docker.com/engine/reference/builder/#copy---link

Zelldon: Ok this is not what I expected but OK :D

oleschoenburg: Yeah, I feel like a lot of the more recent docker features are really subtle and not easy to understand :grimacing:

PR #193, opened 2022-10-12
Allow overriding k8s namespace

oleschoenburg:
This adds a flag similar to `kubeconfig` to override the namespace that will be used by the k8s client.

Depends on #192


go-chaos/internal/k8helper.go
```diff
@@ -70,14 +70,23 @@ func createK8Client(kubeconfig *string) (K8Client, error) {
...
+
+func findKubernetesSettings() KubernetesSettings {
 	//// based on https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go
 	var kubeconfig *string
 	if home := homedir.HomeDir(); home != "" {
 		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
 	} else {
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
 	}
+	namespace := flag.String("namespace", "", "Kubernetes namespace to use")
```
Zelldon, line 86: :x: This is a bit hidden. I would like to use a presistent flag https://stackoverflow.com/a/63498490/2165134 as we do with verbose here https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/cmd/root.go#L53

Zelldon (changes requested):
Thanks :bow: One remark :)

go-chaos/internal/k8helper.go
oleschoenburg: Yeah, that's what I tried first but that introduced a circular dependency from cmd -> internal -> cmd again. I guess that's why the kubeconfig parameter is also defined here and not in cmd?

Zelldon: well what you need to do is then to set a internal property like we do for verbosity see here https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/cmd/root.go#L48

oleschoenburg: Ahh, neat :+1: Do you remember if there was a reason to not do the same for the kubeconfig param?

go-chaos/cmd/root.go
```diff
@@ -39,18 +39,25 @@ var (
...
+
 var rootCmd = &cobra.Command{
 	Use:   "zbchaos",
 	Short: "Zeebe chaos is a chaos experiment tool for Zeebe",
 	Long: `A chaos experimenting toolkit for Zeebe.
     Perfect to inject some chaos into your brokers and gateways.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		internal.Verbosity = Verbose
+		internal.Namespace = Namespace
+		internal.KubeConfigPath = KubeConfigPath
```
Zelldon, line 53: I guess we need also some defaults ?

oleschoenburg: Default for the config path is set here: https://github.com/zeebe-io/zeebe-chaos/pull/193/files#diff-464316c02acb79ad2acc14ad37881cf54b046e8f496271a18a68502f674944b0R79-R88

Zelldon:
works thanks @oleschoenburg :rocket: 

```
[zell go-chaos/ cluster: zeebe-cluster ns:medic-cw-41-966dac308f-benchmark]$ ./zbchaos topology -n medic-cw-41-966dac308f-benchmark-mixed
Node      |Partition 1         |Partition 2         |Partition 3
2         |FOLLOWER (HEALTHY)  |FOLLOWER (HEALTHY)  |FOLLOWER (HEALTHY)
0         |LEADER (HEALTHY)    |FOLLOWER (HEALTHY)  |LEADER (HEALTHY)
1         |FOLLOWER (HEALTHY)  |LEADER (HEALTHY)    |FOLLOWER (HEALTHY)
[zell go-chaos/ cluster: zeebe-cluster ns:medic-cw-41-966dac308f-benchmark]$ ./zbchaos topology
Node      |Partition 1         |Partition 2         |Partition 3
1         |FOLLOWER (HEALTHY)  |LEADER (HEALTHY)    |FOLLOWER (HEALTHY)
2         |FOLLOWER (HEALTHY)  |FOLLOWER (HEALTHY)  |FOLLOWER (HEALTHY)
0         |LEADER (HEALTHY)    |FOLLOWER (HEALTHY)  |LEADER (HEALTHY)

```

PR #201, opened 2022-10-17
feat: add `backup take` and `backup wait` commands

oleschoenburg:
This adds two commands that allow taking a backup and waiting for a backup to complete or fail

### DIFF
--- a/go-chaos/cmd/backup.go
+++ b/go-chaos/cmd/backup.go
@@ -15,7 +15,10 @@
 package cmd
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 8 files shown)


oleschoenburg:
@npepinpe ping? :)

go-chaos/cmd/backup.go
```diff
@@ -0,0 +1,140 @@
...
+
+var (
+	backupId string
+)
+
+func init() {
+	rootCmd.AddCommand(backupCommand)
+
+	backupCommand.AddCommand(takeBackupCommand)
+	takeBackupCommand.Flags().StringVar(&backupId, "backupId", strconv.FormatInt(time.Now().UnixMilli(), 10), "optionally specify the backup id to use, uses the current timestamp by default")
```
npepinpe, line 38: :wrench: Unless we plan to have commands without the flag, it could simply be on the general `backupCommand`, and the child commands will inherit it I think.

go-chaos/cmd/backup.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func takeBackup(cmd *cobra.Command, args []string) error {
+	k8Client, err := internal.CreateK8Client()
+	if err != nil {
+		panic(err)
+	}
+
+	port := 9600
+	closePortForward, err := k8Client.GatewayPortForward(port, port)
```
npepinpe, line 68: :thought_balloon: I think almost everywhere we act the same (e.g. panic when we can't port forward). Wondering if it's useful to have it return an error at all, or if we should just have a function `MustForwardGatewayPort` which returns either a close function or explodes anyway.

npepinpe (approved):
:+1:

PR #210, opened 2022-10-24
feat: add `exporting pause` and `exporting resume` commands

oleschoenburg:


go-chaos/internal/pods.go
```diff
@@ -202,20 +202,20 @@ func (c K8Client) RestartPod(podName string) error {
 // GatewayPortForward creates a port forwarding to a zeebe gateway with the given port
 // https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/tunnel.go#L187-L196
 // https://github.com/kubernetes/client-go/issues/51#issuecomment-436200428
-func (c K8Client) GatewayPortForward(localPort int, remotePort int) (func(), error) {
+func (c K8Client) GatewayPortForward(localPort int, remotePort int) func() {
```
npepinpe, line 205: :wrench: A widespread Go convention is to prefix functions which panic instead of returning error with `Must`, e.g.

```go
func Compute() error {
  // ...
}

func MustCompute() {
  if err := Compute(); err != nil {
    panic(err)
  }
}
```

(not that you need two functions, that was just an example)

npepinpe (approved):
:+1:

PR #221, opened 2022-11-09
feat: commands use provided credentials to connect to the cluster under test

### LINKED ISSUE #220: zbchaos can't connect to clusters over TLS
zbchaos builds a zeebe client with `UsePlaintextConnection` set to true. This makes most zbchaos commands unusable for SaaS clusters.

deepthidevaki:
Previously the client created inside a command to connect to a cluster was always using `UsePlaintextConnection=true`. This would not work with the clusters in ultrachaos. Testbench already passes the credentials via variables. This PR uses that credentials to access the cluster under test.

Manually tested this by running `zbchaos restart --namespace......` against a cluster running in ultrachaos. Also tested against a cluster in our benchmark cluster which requires insecure connection. Both works. Have not tested the zbchaos worker yet, because that would require deploying a new image with this changes. I will test it after merging this PR. 

closes #220

### DIFF
--- a/go-chaos/cmd/root.go
+++ b/go-chaos/cmd/root.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/camunda/zeebe/clients/go/v8/pkg/zbc"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


oleschoenburg (approved):
Nice improvement, thanks @deepthidevaki :+1:

Zelldon:
Thank you both :bow:

PR #229, opened 2022-11-14
Ignore dependabot PR's

Zelldon:
Previously dependabot PR's always fail, since the zbchaos docker build fails. The action has no access to the related secrets when dependabot creates an PR.


oleschoenburg (approved):
Thanks :+1:

PR #231, opened 2022-11-15
Improve current state

Zelldon:
* Move some functions, for better discoverability
* Add new tests
* Support network patch for saas

Can be seen as preparation for working on disconnect gateways

### DIFF
--- a/go-chaos/internal/network.go
+++ b/go-chaos/internal/network.go
@@ -27,21 +27,10 @@ import (
 
 func (c K8Client) ApplyNetworkPatch() error {
 
-	// todo support cloud
-	listOptions := metav1.ListOptions{
-		LabelSelector: "app=camunda-platform",
-	}
-
-	statefulSetList, err := c.Clientset.AppsV1().StatefulSets(c.GetCurrentNamespace()).List(context.TODO(), listOptions)
+	statefulSet, err := c.GetZeebeStatefulSet()
 	if err != nil {
 		return err
 	}
-
-	if len(statefulSetList.Items) <= 0 {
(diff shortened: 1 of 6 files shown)


oleschoenburg (approved):
Looks good, thanks for cleaning up a lot of the stuff I've just left here :sweat_smile: :rocket:

go-chaos/internal/saas.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+)
+
+func (c K8Client) PauseReconciliation() error {
```
npepinpe, line 29: :question: Is there something we could link as documentation for how reconciliation works?

oleschoenburg: https://github.com/camunda-cloud/zeebe-controller-k8s#turning-the-controller-off

go-chaos/internal/network.go
```diff
@@ -27,21 +27,39 @@ import (
 
 func (c K8Client) ApplyNetworkPatch() error {
```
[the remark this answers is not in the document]
Zelldon: Jep right now it is only this I think the other stuff is inside the container like disconnecting

go-chaos/internal/saas.go
Zelldon: Thanks I added it :+1:

PR #232, opened 2022-11-16
Introduce new SaaS flag

Zelldon:
Introduce a new saas flag to simplify code and make it clearer when to use what. This allows to avoid to do multiple calls, like first checking whether brokers pods with self-managed label are deployed and then with SaaS label. With this approach we can know in which environment we are running.

Adjust the tests to also use the CRD to detect the environment.

### DIFF
--- a/go-chaos/internal/deployment.go
+++ b/go-chaos/internal/deployment.go
@@ -26,7 +26,7 @@ import (
 func (c K8Client) getGatewayDeployment() (*v12.Deployment, error) {
 
 	listOptions := metav1.ListOptions{
-		LabelSelector: getGatewayLabels(c.SaaSEnv),
+		LabelSelector: c.getGatewayLabels(),
 	}
 	deploymentList, err := c.Clientset.AppsV1().Deployments(c.GetCurrentNamespace()).List(context.TODO(), listOptions)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


go-chaos/internal/deployment.go
```diff
@@ -24,28 +24,19 @@ import (
...
-		if err != nil {
-			return nil, err
-		}
-
-		// here it is currently hard to distingush between not existing and embedded gateway;
-		// since we don't use embedded gateway in our current chaos setup I would not support it right now here
-		if deploymentList == nil || len(deploymentList.Items) <= 0 {
-			return nil, errors.New(fmt.Sprintf("Expected to find standalone gateway deployment in namespace %s, but none found!", c.GetCurrentNamespace()))
-		}
+		return nil, errors.New(fmt.Sprintf("Expected to find standalone gateway deployment in namespace %s, but none found!", c.GetCurrentNamespace()))
```
npepinpe, line 39: :wrench: As per the comment, I would add a small bit after saying we don't support embedded gateway, e.g.

```go
return nil, errors.New(fmt.Sprintf("Expected to find standalone gateway deployment in namespace %s, but none found! The embedded gateway is not supported", c.GetCurrentNamespace()))
```

npepinpe (approved):
:+1: 

Just one small suggestion, nothing big

go-chaos/internal/deployment.go
Zelldon: Good point thanks

Issue #226, opened 2022-11-11
I can deploy models via Zbchaos


Zelldon:
related to https://github.com/zeebe-io/zeebe-chaos/issues/139

To replace scripts:

- [ ] deploy-different-versions.sh
- [ ]  deploy-model.sh
- [ ]  deploy-specific-model.sh

Issue #160, opened 2022-08-03
Improve Deployment Distribution experiment


Zelldon:
See https://zeebe-io.github.io/zeebe-chaos/2022/08/02/deployment-distribution we found out that DMN is not deployed during the experiment we should add that to cover that as well. 

related to #61

PR #240, opened 2022-11-17
Deploy models via zbchaos

Zelldon:
## Details
Add new sub command `deploy process` to deploy process models, either specified by the user
via `--processModelPath` or if not given it will deploy multiple versions of default models (BPMN and DMN).

The user is able to specify the count of versions which should be deployed for each.
This allows experimenting with deployment distribution.

Replacement of:

-  deploy-different-versions.sh
-  deploy-model.sh
-  deploy-specific-model.sh

closes https://github.com/zeebe-io/zeebe-chaos/issues/226
closes https://github.com/zeebe-io/zeebe-chaos/issues/160

### Example Usage:

```
$ ./zbchaos deploy process --verbose         
Connecting to zell-chaos
Deploy 10 versions of different type of models.
Deployed [2/10] versions.
Deployed [4/10] versions.
Deployed [6/10] versions.
Deployed [8/10] versions.
Deployed [10/10] versions.
Deployed different process models of different types and versions to zeebe!
```

bpmn
dmn

### DIFF
--- a/go-chaos/cmd/deploy.go
+++ b/go-chaos/cmd/deploy.go
@@ -1,0 +1,81 @@
+// Copyright 2022 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/zeebe-io/zeebe-chaos/go-chaos/internal"
+)
+
+func init() {
+	rootCmd.AddCommand(deployCmd)
+
+	deployCmd.AddCommand(deployProcessModelCmd)
+
+	deployProcessModelCmd.Flags().StringVar(&processModelPath, "processModelPath", "",
+		"Specify the path to a BPMN process model, which should be deployed.")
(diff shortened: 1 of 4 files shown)


Zelldon:
I also tried to improve our test coverage related to these functionality, I wanted to verify for example that X deployments are send or deployment X is created. Unfortunately I'm not able to fake/mock the Zeebe client because of certain issues, see https://github.com/camunda/zeebe/issues/11044 (https://github.com/camunda/zeebe/issues/11045)

npepinpe (approved):
Regarding testing the Zeebe client, I don't understand why gomock wouldn't work :thinking: 

Anyway, one way to circumvent this is to have your code not use a ZeebeClient but instead use an interface type you describe which specifies how to deploy, e.g. `MyDeployer`. Since Go uses structural typing, any type which implements the function will automatically implement the type, so it will work with the Zeebe client and your own mock. You have to keep the interface in sync or adapter for the Zeebe client though, but yeah. I guess it's do-able :shrug: Hope I'm clear :sweat_smile:

Zelldon:
Hey @npepinpe thanks for the review :bow: 

Haven't tried yet gomock, was also a bit reluctant to use something where I have to generate something (or use a tool).

I also thought to change my functions to only use parts which might be better intercepted. You input gave also a good trigger regarding the own interface which I tried out, but without success. You can take a look at this here https://github.com/zeebe-io/zeebe-chaos/commit/d4cfdc4e584565e55d810adb1599cef062571d99

What we can see is that sure I can create such an interface, but still i have an issue that the ZeebeCLient returns a struct which is not what I want to use or return in that case (since I can't intercept the call or control anything then) Maybe I'm just too stupid for this :see_no_evil: 

extract

I feel a bit lost here. :smiling_face_with_tear:

Zelldon:
> I think you'd have to workaround this by essentially creating interfaces for everything you need, and ensuring either the Zeebe types match those (not just the client, but the returned structs for the complete execution path), or writing adapters yourself. I don't have time to check it out today, and to be honest I think a fake client is good idea anyway for us, but if you'd like we can take a look next week.

I thought this as well and thought but why I use then anyway this client at all :sweat_smile: than I can also create a complete new client :laughing:

Issue #244, opened 2022-11-22
Create instance with specific version


Zelldon:
We extend the verify steady-state command to specify a version, such that we can create an instance with a version. If version is specified no deployment should be deployed. 

This allows to execute the deployment distribution experiment.

PR #246, opened 2022-11-22
Create process instance for specific version

Zelldon:
Allows creating a process instance for a specific version and BPMN process id via the verify steady-state command.
This is useful for the deployment distribution command.

Furthermore, I also created the first version of a Zeebe Fake client, in order to intercept values which are set on the client requests. This allows us to create more unit tests where we can verify whether the right properties and values are set.

Added some new tests for the process instance creation via version and BPMN process id.

closes https://github.com/zeebe-io/zeebe-chaos/issues/244

-----

**Example:**

Deploy model
```sh
$ ./zbchaos deploy process -v
Connecting to zell-chaos
Running experiment in self-managed environment.
Deploy 10 versions of different type of models.
Deployed [2/10] versions.
Deployed [4/10] versions.
Deployed [6/10] versions.
Deployed [8/10] versions.
Deployed [10/10] versions.
Deployed different process models of different types and versions to zeebe!

```

Non-verbose: Start instance for a specific version
```
$ ./zbchaos verify steady-state --version 10 --bpmnProcessId multiVersion
The steady-state was successfully verified!
```

We can see in operate:

execution

Verbose: Start instance for a specific version

```sh
$ ./zbchaos verify steady-state --version 10 --bpmnProcessId multiVersion -v
Connecting to zell-chaos
Running experiment in self-managed environment.
Create process instance with BPMN process ID multiVersion and version 10 [variables: '', awaitResult: false]
Created process instance with key 4503599627370497 on partition 2, required partition 1.
Created process instance with key 6755399441055745 on partition 3, required partition 1.
Created process instance with key 2251799813685299 on partition 1, required partition 1.
The steady-state was successfully verified!
```

execution2

### DIFF
--- a/go-chaos/cmd/verify.go
+++ b/go-chaos/cmd/verify.go
@@ -15,12 +15,9 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"time"
 
-	"github.com/camunda/zeebe/clients/go/v8/pkg/commands"
-	"github.com/camunda/zeebe/clients/go/v8/pkg/zbc"
 	"github.com/spf13/cobra"
 	"github.com/zeebe-io/zeebe-chaos/go-chaos/internal"
 )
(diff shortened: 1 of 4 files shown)


go-chaos/cmd/verify.go
```diff
@@ -33,6 +36,8 @@ func init() {
 	verifySteadyStateCmd.Flags().StringVar(&processModelPath, "processModelPath", "", "Specify the path to a BPMN process model, which should be deployed and an instance should be created of.")
 	verifySteadyStateCmd.Flags().StringVar(&variables, "variables", "", "Specify the variables for the process instance. Expect json string.")
 	verifySteadyStateCmd.Flags().BoolVar(&awaitResult, "awaitResult", false, "Specify whether the completion of the created process instance should be awaited.")
+	verifySteadyStateCmd.Flags().IntVar(&version, "version", -1, "Specify the version for which the instance should be created.")
```
oleschoenburg, line 39: :question: Is there a way to mark these flags as optional and depending on each other? I fear that as we are adding more options, it's becoming less clear what flags need to be set and what their dependencies are. For example, it's not clear that when `version` is set, `bpmnProcessId` is required and `processModelPath` has no effect.

oleschoenburg (approved):
LGTM :+1: 

Good idea to start mocking zeebe client!

go-chaos/cmd/verify.go
Zelldon: I will follow up on this https://github.com/zeebe-io/zeebe-chaos/issues/249

PR #255, opened 2022-11-24
Refactor Backend

Zelldon:
* I extracted code from connect and disconnect to  a new package, backend. 
* The code was refactored such it returns errors, instead of doing panic. Panic is done in commands currently.

I thought I can reuse the code and call it in an extra worker, but the problem I see is the argument parsing. I would need to do it for everything again somehow.

I realized that the for each command an `RunE` can be set instead of `Run` which allows to return an error. This would be useful for the zbchaos worker, since when now executing a command it would return an error instead of panicking. 

Wdyt @oleschoenburg does this look like a solution for you? Would like to get your opinion. I did that as example in the last commit.

I need to run this as cli locally to try it out how it behaves then.

### DIFF
--- a/go-chaos/Dockerfile
+++ b/go-chaos/Dockerfile
@@ -6,9 +6,7 @@ WORKDIR /app
 COPY --link go.mod go.sum .
 RUN go mod download
 
-COPY --link cmd/ ./cmd
-COPY --link internal ./internal
-COPY --link main.go ./
+COPY --link . ./
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-s -w" -o zbchaos main.go
 
 FROM scratch
--- a/go-chaos/backend/connection.go
+++ b/go-chaos/backend/connection.go
@@ -1,0 +1,107 @@
+// Copyright 2022 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/zeebe-io/zeebe-chaos/go-chaos/internal"
+	v1 "k8s.io/api/core/v1"
+)
+
+func ConnectBrokers() error {
+	k8Client, err := internal.CreateK8Client()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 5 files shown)


Zelldon:
so if an error is return on `RunE`, The error will also printed by cobra and the help

```
 ./zbchaos disconnect brokers -v
....
Execute ["apt" "-qq" "update"] on pod zell-chaos-zeebe-2
Error: unable to upgrade connection: container not found ("zeebe")
Usage:
  zbchaos disconnect brokers [flags]

Flags:
      --broker1NodeId int        Specify the nodeId of the first Broker (default -1)
      --broker1PartitionId int   Specify the partition id of the first Broker (default 1)
      --broker1Role string       Specify the partition role [LEADER, FOLLOWER] of the first Broker (default "LEADER")
      --broker2NodeId int        Specify the nodeId of the second Broker (default -1)
      --broker2PartitionId int   Specify the partition id of the second Broker (default 2)
      --broker2Role string       Specify the partition role [LEADER, FOLLOWER] of the second Broker (default "LEADER")
  -h, --help                     help for brokers
      --one-direction            Specify whether the network partition should be setup only in one direction (asymmetric)

Global Flags:
      --audience string       connect using the given client secret
  -c, --clientId string       connect using the given clientId
      --clientSecret string   connect using the given client secret
      --jsonLogging           json logging output
      --kubeconfig string     path the the kube config that will be used
  -n, --namespace string      connect to the given namespace
  -v, --verbose               verbose output

unable to upgrade connection: container not found ("zeebe")
```

oleschoenburg (approved):
I think it's definitely better than crashing everywhere. However, it still relies on our diligence to not panic anywhere and also to not call any functions that might panic, right?

I'm approving now since the refactorings look good to me and it's definitely a step in the right direction (and it's unclear to me if it's even possible to improve from here)

Zelldon:
>  However, it still relies on our diligence to not panic anywhere and also to not call any functions that might panic, right?

I guess we can scan the code for panic's.

BTW I just thought about the whole thing again and now I'm not sure anymore whether it is a problem of the panic, since even if it fails the worker would restart and we would retry the job might be also nice to have. 

This is actually also what we had before https://github.com/zeebe-io/zeebe-chaos/blob/main/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/chaosMain.kt#L142-L150 and If I remember correctly also decided for, because if we would fail the job we would get incidents which we wanted to avoid afaik. 

Be aware the task is canceled after certain timeout anyway by the process. 

Right now I'm not sure whether it make sense to refactor it more. I guess I need to try out first how it behaves as worker.

Any additional thoughts @oleschoenburg ?

oleschoenburg:
Yeah, the crash and restart behavior is what I expected and I think I've seen it happen while testing the backup qa stuff. I'm fine with either option, there's not that much difference IMO.

Zelldon:
Thanks :+1: ok I will bring back the panic on the disconnect if an error exists and merge the rest of the refactoring.

The other commands I will keep with the panics for now and when I have time refactor more to move to the backend package and return errors and only panic in the commands, is a bit cleaner.

PR #256, opened 2022-11-24
Deploy chaos process models

Zelldon:
Allows to bundle the chaos experiments with zbchaos and deploy them via `zbchaos deploy chaos` command.

This simplifies the testing of them and we can use that in the k8 deployment to first deploy the models and then start the worker.

### DIFF
--- a/go-chaos/backend/clients.go
+++ b/go-chaos/backend/clients.go
@@ -1,0 +1,20 @@
+package backend
+
+import (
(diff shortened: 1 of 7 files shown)


oleschoenburg (approved):
Love it, `zbchaos deploy chaos` reminds me of 

Release the kraken

Zelldon:
haha nice :D

PR #257, opened 2022-11-24
New tests for worker

Zelldon:
* Move logic of worker into own package.
  * For better encapsulation and testability
* Write new tests for job worker/handler

### DIFF
--- a/go-chaos/worker/chaos_worker.go
+++ b/go-chaos/worker/chaos_worker.go
@@ -1,0 +1,62 @@
+package worker
+
+import (
+	"context"
+	"time"
+
+	"github.com/camunda/zeebe/clients/go/v8/pkg/entities"
+	"github.com/camunda/zeebe/clients/go/v8/pkg/worker"
+)
+
+var CommandRunner func([]string, context.Context) error
+
+type ChaosProvider struct {
+	Path      string
+	Arguments []string
+	Timeout   int64
+}
+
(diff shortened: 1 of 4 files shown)


go-chaos/worker/chaos_worker.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/camunda/zeebe/clients/go/v8/pkg/entities"
+	"github.com/camunda/zeebe/clients/go/v8/pkg/worker"
+)
+
+var CommandRunner func([]string, context.Context) error
```
oleschoenburg, line 25: :question: This is a globally shared, mutable variable, right? I'm not sure if it's idiomatic in go but I'd have expected a `ChaosWorker` struct that encapsulates the handler function. Just to get rid of the global var.

oleschoenburg (approved):
:+1: I like the refactoring, just one minor comment

go-chaos/worker/chaos_worker.go
Zelldon: Yeah I was not 100% sure where to put it then. But I guess I need a wrapper around the handler to insert more inputs.

Zelldon: Good point @oleschoenburg I fixed it :)

PR #258, opened 2022-11-25
Worker for read chaos experiments

### LINKED ISSUE #236: I can read all chaos experiments via zbchaos
Similar to the kotlin worker the zbchaos worker need to be able to publish the chaos experiments which we run

This is necessary to migrate to zbchaos

Zelldon:
Replacement of old Kotlin worker which has read the experiments, see https://github.com/zeebe-io/zeebe-chaos/blob/main/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/ReadChaosExperimentsHandler.kt

The Camunda cloud experiments are now bundled into zbchaos, allowing it to be next to the code and not fail experiments when changing something in the repo. This means experiments are versioned with zbchaos now. 

Added new tests and extended the fake job client to write better tests.

closes #236

### DIFF
--- a/go-chaos/cmd/worker.go
+++ b/go-chaos/cmd/worker.go
@@ -25,7 +25,8 @@ import (
 	worker "github.com/zeebe-io/zeebe-chaos/go-chaos/worker"
 )
 
-const jobType = "zbchaos"
+const jobTypeZbChaos = "zbchaos"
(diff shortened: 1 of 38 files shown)


oleschoenburg (approved):
We are getting there... :partying_face:

PR #259, opened 2022-11-30
fix: Do not wait for all pods to be ready in dataloss recovery

deepthidevaki:
When recovering after dataloss, there might be other brokers which are not yet recovered. So we cannot wait for all brokers to be ready.


Zelldon (approved):
:+1: :)

PR #260, opened 2022-11-30
ci: build a snapshot image of zbchaos

deepthidevaki:
This is useful for testing and no need to wait until the release. The image is not tagged as SNAPSHOT but with the commit sha.


Zelldon (approved):
:+1: I guess this is ok

PR #264, opened 2022-12-02
Experiments need to start with lower letter

Zelldon:
Set json annotation to marshal the experiments into correct key, with small letters.


Zelldon:
This causes right now several incidents in our QA runs.

PR #277, opened 2022-12-07
fix: for readiness check both the phase and ready status

deepthidevaki:
Observed that the ready status = true when the pod is terminating.

### DIFF
--- a/go-chaos/internal/pods.go
+++ b/go-chaos/internal/pods.go
@@ -121,32 +121,33 @@ func (c K8Client) RestartPodWithGracePeriod(podName string, gracePeriodSec *int6
 }
 
 func (c K8Client) AwaitReadiness() error {
-	retries := 0
-	maxRetries := 300 // 5 * 60s
+	return c.AwaitReadinessWithTimeout(5 * time.Minute)
+}
+
+func (c K8Client) AwaitReadinessWithTimeout(timeout time.Duration) error {
+	timedOut := time.After(timeout)
+	ticker := time.Tick(1 * time.Second)
+
+	// Keep checking until we're timed out
 	for {
-		if retries >= maxRetries {
-			return errors.New("Awaited readiness of pods in namespace %s, but timed out after 30s")
-		}
-		if retries > 0 {
-			time.Sleep(1 * time.Second)
-		}
+		select {
+		case <-timedOut:
+			return errors.New(fmt.Sprintf("Awaited readiness of pods in namespace %v, but timed out after %v", c.GetCurrentNamespace(), timeout))
+		case <-ticker:
+			brokersAreRunning, err := c.checkIfBrokersAreRunning()
+			if err != nil {
+				LogVerbose("Failed to check broker status. Will retry. %v", err)
+			}
+			gatewaysAreRunning, err := c.checkIfGatewaysAreRunning()
+			if err != nil {
+				LogVerbose("Failed to check gateway status. Will retry. %v", err)
+			}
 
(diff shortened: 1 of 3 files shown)


Zelldon (commented):
I would feel better if we have a test for that. Could you take a look at this?

Like adding a test where you setup a pod, with pod status Running, Failed etc. and verify the result of that check function?


Here you find the functions which creates the pod https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/internal/helper_test.go#L73-L80 in the fake k8 environment

Here https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/internal/pods_test.go I would see the test like

* Test_ShouldReturnTrueWhenBrokerPodsAreReady
* Test_ShouldReturnFalseWhenOneBrokerPodIsNotReady

deepthidevaki:
@Zelldon Do you know how I can mock a respose for "get pods" to return ready or not ready pods? When creating a pod using the fakeclient we are only providing the spec of the pod. We cannot set the status.

Zelldon:
@deepthidevaki you can define the status here in the struct :) https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/internal/helper_test.go#L73-L80 Maybe create another function which creates the pod with a status.

deepthidevaki:
> @deepthidevaki you can define the status here in the struct :) https://github.com/zeebe-io/zeebe-chaos/blob/main/go-chaos/internal/helper_test.go#L73-L80 Maybe create another function which creates the pod with a status.

Aah, Right. I tried to add it in the Spec :see_no_evil:

go-chaos/internal/pods.go
```diff
@@ -121,32 +121,33 @@ func (c K8Client) RestartPodWithGracePeriod(podName string, gracePeriodSec *int6
...
-		gatewaysAreRunning, err := c.checkIfGatewaysAreRunning()
-		if err != nil {
-			return err
-		}
+	// Keep checking until we're timed out
+	for {
+		select {
+		case <-timedOut:
+			return errors.New(fmt.Sprintf("Awaited readiness of pods in namespace %v, but timed out after %v", c.GetCurrentNamespace(), timeout))
+		case <-ticker:
```
Zelldon, line 136: :+1: Yeah I thought also last week that we can write this better with channels thanks for improving this :rocket:

go-chaos/internal/pods_test.go
```diff
@@ -282,8 +283,50 @@ func Test_ShouldReturnTrueWhenBrokersAreReady(t *testing.T) {
...
+	require.NoError(t, err)
+	k8Client.CreateBrokerPodsWithStatus(t, selector, "broker1", v1.PodRunning, true)
+	k8Client.CreateBrokerPodsWithStatus(t, selector, "broker2", v1.PodRunning, false)
+
+	gatewaySelector, err := metav1.ParseToLabelSelector(getSelfManagedGatewayLabels())
+	require.NoError(t, err)
+	k8Client.CreateDeploymentWithLabelsAndName(t, gatewaySelector, "gateway")
+
+	// when
+	err = k8Client.AwaitReadinessWithTimeout(2 * time.Second)
```
Zelldon, line 307: :wrench: Could we reduce the timeout otherwise the tests take a bit long

Zelldon (approved):
Thank you very much for adding additional tests @deepthidevaki :+1: :rocket: 

Please have a look at my comments, maybe you can do that as follow up. I will merge it anyway, since I'm currently testing the QA and this helps here.

PR #285, opened 2022-12-08
Multiple zbchaos fixes

Zelldon:
During deploying and running zbchaos with our QA runs we encountered several issues which I fix in this PR.

## Env vars overwrite values

When deploying zbchaos and on the zbchaos worker we need to use different names for the environment variables.

When we using the normal ZEEBE_ environment variables we run into conflicts with using multiple zeebe clients, the env vars will always overwrite direct values.

## Plaintext

When setting the credentials provider we also have to set the UsePlainText to false.

## No credentials provider for inside the cluster

We need no credentials when running against the target cluster since we port forward to the gateway. This means we can remove the credentials usage.

## Await result

The awaiting of the process instances doesn't require the partition id check. Moreover, this is problematic for things like message correlation since a message is correlated only once to a process. 

Remove the partition id check, when `awaitResult` flag is enabled.


Zelldon:
I will go ahead and merge these changes since these clearly fixes some bugs in running our QA

PR #289, opened 2022-12-09
Encapsulate all commands

Zelldon:
In order to not run into issues with global variables reused and reset during worker execute, we encapsulated all commands and make it available via `NewCMD` so everytime we have to create a new command to execute zbchaos in the workers. 

This allows us to always use a clean context.


Zelldon:
@oleschoenburg with the last commit I was also able to move the namespace and kubeconfig to the flags struct

oleschoenburg (approved):

PR #291, opened 2022-12-09
Use flags pointer

Zelldon:
In order to be able to see value changes by cobra we need to pass pointers around


Zelldon:
* https://github.com/zeebe-io/zeebe-chaos/pull/295
* https://github.com/zeebe-io/zeebe-chaos/pull/296

PR #363, opened 2023-05-19
Verify creation and completion of X process instances

Zelldon:
Add new feature which allows us to create variable count of process instances, with configurable bpmn process id etc.
Furthermore a feature to verify that variable amount of jobs can be completed.

These both features allowed to run a chaos day https://zeebe-io.github.io/zeebe-chaos/2023/05/19/Continuing-SST-Partitioning-toggle

Were we wanted to create multiple process instances, run an chaos action and complete the created instances later. This includes with verifying the exact amount of instances/jobs.


go-chaos/cmd/verify.go
```diff
@@ -80,9 +80,68 @@ Process instances are created until the required partition is reached.`,
 		},
 	}
 
+	var verifyInstanceCount = &cobra.Command{
+		Use:   "instance-count",
+		Short: "Verify the instance creation count",
+		Long:  `Verifies that an specific count of process instances from a specific process model can be created.`,
```
deepthidevaki, line 86:
```suggestion
		Long:  `Verifies that a specific count of process instances from a specific process model can be created.`,
```

go-chaos/cmd/verify.go
```diff
@@ -80,9 +80,68 @@ Process instances are created until the required partition is reached.`,
...
+				BpmnProcessId: flags.bpmnProcessId,
+				Version:       int32(flags.version),
+				AwaitResult:   false,
+				Variables:     flags.variables,
+			})
+			ensureNoError(err)
+			err = internal.SendCountOfCommands(internal.ZCCommandSender(processInstanceCreator), int32(flags.instanceCount), time.Duration(flags.timeoutInSec)*time.Second)
+			ensureNoError(err)
+
+			internal.LogInfo("The steady-state was successfully verified!")
```
deepthidevaki, line 109: :wrench: Improve log message to indicate x process instances were created.

go-chaos/cmd/verify.go
```diff
@@ -80,9 +80,68 @@ Process instances are created until the required partition is reached.`,
...
+			defer zbClient.Close()
+
+			jobCompleter, err := internal.CreateJobCompleter(zbClient, internal.JobCompleteOptions{
+				JobType: flags.jobType,
+			})
+			ensureNoError(err)
+			err = internal.SendCountOfCommands(jobCompleter, int32(flags.jobCount), time.Duration(flags.timeoutInSec)*time.Second)
+			ensureNoError(err)
+
+			internal.LogInfo("The steady-state was successfully verified!")
```
deepthidevaki, line 136: :wrench: Same here. Improve log message to indicate x jobs were completed.

deepthidevaki (approved):
Thanks. 

:thought_balloon: There might be some cases where this way of counting is not accurate. For example, due to leader change/restarts jobs can timeout and get activated again. So same job will be counted twice, right? Depending on the purpose of this feature, this might be acceptable.

PR #364, opened 2023-05-23
fix: remove unnecessary I/O mapping

npepinpe:
Removes an unnecessary I/O mapping when an experiment fails. This triggered an incident in Testbench because there is no variable source. It also doesn't seem to me like this was required at all.

Other changes were due to me using a much newer version of the modeler.


deepthidevaki (approved):
Thanks.

How did this worked before? There were test failures in past :thinking:

npepinpe:
Was this variable supposed to be set? Because I couldn't find references to it elsewhere or in further steps - but I'm not very familiar with the chaos BPMNs.

deepthidevaki:
> Was this variable supposed to be set? Because I couldn't find references to it elsewhere or in further steps - but I'm not very familiar with the chaos BPMNs.

If it was supposed to set, it should have more meaningful names. This looks like a left-over. So I guess it can be removed.

Zelldon:
FYI End events didn't allow output mappings before. Was implemented https://github.com/camunda/zeebe/pull/10618

PR #365, opened 2023-05-30
feat: new encrypted Kubernetes configuration file

multani:
The file is encrypted using sops and a GCP KMS key.


Zelldon (approved):
:heart: Thank you so much for that @multani :rocket:

PR #374, opened 2023-06-29
fix: only delete pvc

deepthidevaki:
Reclaim policy of pvc is "Delete", which means it will be automatically deleted when PV is deleted and it is no longer attached to a pod. 

When we delete both PVC and PV together, PV is not getting deleted. We don't know yet why this happens. But we (with @multani ) verified that when we delete PVC, the PV is also getting deleted.

https://github.com/camunda-cloud/team-sre/issues/253#issuecomment-1609259799

### DIFF
--- a/go-chaos/internal/volume.go
+++ b/go-chaos/internal/volume.go
@@ -33,11 +33,6 @@ func (c K8Client) DeletePvcOfBroker(podName string) error {
 		return err
 	}
 
-	LogInfo("Deleting PV %s", pvc.Spec.VolumeName)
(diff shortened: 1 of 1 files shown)


multani:
> When we delete both PVC and PV together, PV is not getting deleted. We don't know yet why this happens. But we (with @multani ) verified that when we delete PVC, the PV is also getting deleted.

Moreover, when we explicitly delete the PV, the underlying disk on GCP doesn't get deleted automatically (I'm trying to understand the mechanism here).

As mentioned by @deepthidevaki , the reclaim policy configured on the PVC will automatically delete the related PV _and_ the underlying GCP disk when the PVC is deleted.

Zelldon:
Thanks 👍 I was already wondering why we delete PV explicitly 😄

Zelldon (approved):
🚀

PR #388, opened 2023-08-01
fix(setup): runAs root to install tooling

megglos:
Since https://github.com/camunda/zeebe/pull/13418 got merged the zeebe container runs with an unprivileged `zeebe` user.

As the chaos tooling makes use of apt to install some tools needed to e.g. stress the cpu or modify ip routes we need to also overwrite the `runAsUser` to root to make that possible still.

Unfortnately the k8s exec API is not offering a way to override the user see https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/exec/exec.go#L104

The first commit resolves an issue that lead to test failure, see https://github.com/testcontainers/testcontainers-go/issues/1359#issuecomment-1657098312


oleschoenburg (approved):
Thanks for finding a fix :+1:

PR #396, opened 2023-08-22
fix: publish message with 1h ttl

korthout:
Messages were published with a TTL of 5 minutes, which means the engine buffers them for 5 minutes. The buffered message is then expected to correlate to a process instance's event later in the chaos experiment.

However, in some cases, the chaos experiment takes longer than these 5 minutes. When the message is expired, the chaos experiment can no longer succeed even though it retries for an hour. For example, this occurred in the action:
- Should be able to create a process instance and await the message correlation

Changing the TTL of the published message fixes this problem by buffering the message for the same duration as the retry period.

### DIFF
--- a/go-chaos/cmd/publish.go
+++ b/go-chaos/cmd/publish.go
@@ -48,7 +48,7 @@ func AddPublishCmd(rootCmd *cobra.Command, flags *Flags) {
 
 			internal.LogVerbose("Send message '%s', with correaltion key '%s' (ASCII: %d) ", flags.msgName, correlationKey, int(correlationKey[0]))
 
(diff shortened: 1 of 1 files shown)


Zelldon (approved):
Awesome! Thank you very much @korthout that you looked into it and fixed it! Great stuff! 🚀 

I was think whether there was any explanation why it was 5 minute, so I searched the git history. I just found this https://github.com/zeebe-io/zeebe-chaos/commit/f9adf479f1401cf1d9adf9a57c4d7ecd5e1c8eaa I remember before I had no TTL which is in go per default zero (I think) and also failed the experiments. With 5 minutes I think it most of the time just worked. But yes makes sense to me to increase this a bit more to be more reliable 👍🏼

PR #397, opened 2023-08-24
Add release workflow

Zelldon:
I don't want to follow the manual guide anymore.

This PR adds a new workflow which allows to manually trigger zbchaos release, following the guide documented in the README https://github.com/zeebe-io/zeebe-chaos/blob/zbchaosv1.3.0/go-chaos/README.md

What is missing to autodeploy the zbchaos docker image, after it is created by the other github action. But step by step.

Resources:

 * https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows
 * https://github.com/zeebe-io/zeebe-chaos/blob/zbchaosv1.3.0/go-chaos/README.md
 * https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,38 @@
+name: Release zbchaos
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release and tag name'     
+        required: true
+        default: 'zbchaosv0.X.0'
+
+env:
+  GO_VERSION: 1.19
+
+jobs:
+ go-release:
+    name: Run Go CI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: go-chaos
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      RELEASE_VERSION: ${{ github.event.inputs.release }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
+name: Release zbchaos
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release and tag name'     
+        required: true
+        default: 'zbchaosv0.X.0'
```
oleschoenburg, line 8: :wrench: 
```suggestion
        default: 'zbchaos v1.X.0'
```

oleschoenburg (approved):
:rocket:

.github/workflows/release.yaml
Zelldon: @oleschoenburg I think we cant do it because the tag will be created with it, but how about using a `-`.

oleschoenburg: Ah okay yeah. Then we need to change this too I think: https://github.com/zeebe-io/zeebe-chaos/blob/main/.github/workflows/publishZbchaosImage.yml#L24

PR #410, opened 2023-09-21
fix: use different image registry for busybox

### LINKED ISSUE #403: BUG: Don't use busybox image in SaaS
SaaS will block other registries to control what images are deployed.

We need to adjust the dataloss helper, to specific the [initcontainer ](https://github.com/zeebe-io/zeebe-chaos/blob/4dbdaff6ea8f2d44069aea164f50beca0effcab0/go-chaos/internal/dataloss_sim_helper.go#L67-L98) and use `gcr.io/camunda-saas-registry/busybox:1.36.1`

Related discussion https://camunda.slack.com/archives/CSQ2E3BT4/p1694600290842769?thread_ts=1694446510.955439&cid=CSQ2E3BT4

deepthidevaki:
Busybox image is set differently for SaaS and SM. For SM the tag is set to `latest`. But it seems tag latest is not available in SaaS registry. So it is not set to the latest version `1.36.1`. 

closes #403

### DIFF
--- a/go-chaos/internal/dataloss_sim_helper_test.go
+++ b/go-chaos/internal/dataloss_sim_helper_test.go
@@ -1,0 +1,56 @@
+// Copyright 2023 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


rodrigo-lourenco-lopes (approved):
LGTM! 🚀 🚀

go-chaos/internal/dataloss_sim_helper_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 Camunda Services GmbH
```
Zelldon, line 1: 🚀

Zelldon (approved):
Thank you! 🤗

go-chaos/internal/dataloss_sim_helper_test.go
oleschoenburg: I guess that should be 2023 :sweat_smile:

PR #428, opened 2023-10-26
refactor: add error logging for detecting saas env

Zelldon:
We currently silently hide errors on detecting the SaaS environment, this PR should at least log them.


Zelldon:
More an fyi I will go ahead an merge.

megglos (commented):
👍

PR #429, opened 2023-10-26
Automate zbchaos release

Zelldon:
Release is now fully automated, you can either run the `./release` script under `go-chaos/` or use the corresponding github action. 

The release script contains:

 1. determining the release version
 2. building the artifacts
 3. creating a github release
 4. building and pushing docker images
 5. deploying the resources to gke


go-chaos/release.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "Nice, you want to release a new version of zbchaos."
+echo "Checking out main and retrieving most recent tags"
+git checkout main
+git fetch --all
+git pull origin main
+
+lastRelease=$(git tag | sort -Vr | head -n1)
+
+echo "Last release was: $lastRelease"
+read -p "Please type a new release here: " newRelease
```
oleschoenburg, line 23: :x: This won't work when run from GH actions. Maybe allow providing `newRelease` as an environment variable?

Zelldon: Thanks good catch, totally forgot about this 😆

Zelldon: Thanks for checking @oleschoenburg should be fixed

Issue #446, opened 2023-12-06
Ensure we run experiments with worker images using the same version as the target cluster


npepinpe:
**Description**

We need to make sure that when we run experiments, we run them with the same worker version as the target cluster which we want to test.

> Note that we should also do this for the starter eventually, which is the zbchaos image itself, but this is not blocking at the moment.

PR #447, opened 2023-12-06
Use docker image for the workers

Zelldon:
Allow to specify a dockerImageTag as global parameter, to be used by the deployed worker (and maybe later starter, etc.)

This is to make sure that we use the same client version as the deployed/target zeebe cluster.

closes https://github.com/zeebe-io/zeebe-chaos/issues/446


npepinpe:
ups, no bors here I guess :smile: 

I'll merge once tests pass

Zelldon:
I can fix the test later.

PR #450, opened 2023-12-07
Allow restarting and terminating all gateways

npepinpe:


Zelldon (approved):
Thanks 🚀 one step closer 🎸 👨🏼‍🎤

PR #451, opened 2023-12-07
feat: allow running versioned experiments

npepinpe:
This PR allows us to filter the experiments we want to run in a declarative way. Experiments must now be explicitly defined in the [new manifest file embedded](go-chaos/internal/chaos-experiments/camunda-cloud/manifest.yml) in the chaos worker.

The manifest looks like this:

```yaml
experiments:
  - path: broker-dataloss/experiment.json
    clusterPlans:
      - production-s
  - path: deployment-distribution/experiment.json
    clusterPlans:
      - production-s
  # only for testing
  - path: test/experiment.json
    clusterPlans:
      - test
  - path: test/version-experiment.json
    clusterPlans:
      - test
    minVersion: 8.3
    maxVersion: 8.4
```

> [!Note]
> For version bounds, we only check the major and minor version. So 8.3, 8.4, etc. If the target cluster is 8.4.0-SNAPSHOT and the bound is 8.4, then it will match.

The logic when reading the manifest is as follows:

- Given the current target cluster plan X, only experiments who have `X` in their `clusterPlans` list will be picked.
- Given the current target cluster version X, only experiments whose `minVersion >= X` `maxVersion <= X` are picked.
- The `minVersion` and `maxVersion` parameters are completely optional. If you specify none, then the experiment runs on any version.
- If we somehow fail to get the target cluster version, any experiment with version bound is omitted.

The target version is obtained by getting the topology of the target cluster, either from a given namespace, or from the current namespace of the kubeconfig if no namespace was provided.


npepinpe:
It seems this line causes a panic:

```
list, err := c.Clientset.CoreV1().Pods(c.GetCurrentNamespace()).List(context.TODO(), listOptions)
```

npepinpe:
no? it fixed the issue in ci without a cluster

Zelldon:
Wondering whether it makes sense to put it actually in these methods, because then we would also need it in the others right? Maybe it makes more sense after creating the client, whether it is correctly initialized.

go-chaos/worker/chaos_worker.go
```diff
@@ -163,3 +172,28 @@ func HandleReadExperiments(client worker.JobClient, job entities.Job) {
 
 	_, _ = command.Send(ctx)
 }
+
+func getTargetClusterVersion(namespace string) string {
+	k8Client, err := internal.CreateK8Client("", namespace)
+	port := 26500
+	closeFn, err := k8Client.GatewayPortForward(port, port)
```
Zelldon, line 179: Here I mean to check the client

npepinpe: Ah, gotcha :+1: Makes sense

Zelldon (approved):
Awesome thanks @npepinpe 🚀 🎸

PR #452, opened 2023-12-08
Allow injecting polling delay into worker deployment

npepinpe:
**Description**

This PR allows injecting the polling delay into the worker deployment. The polling delay is given via command line as milliseconds because I couldn't be bothered to figure out how to convert to a nicer type from string :upside_down_face: 

This will now allow us to "disable" polling for job push experiments.

It also fixes setting a default value for the worker Docker image tag, as it was previously defaulting to `""` when omitted.

### DIFF
--- a/go-chaos/internal/manifests/worker.yaml
+++ b/go-chaos/internal/manifests/worker.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: worker
-          image: gcr.io/zeebe-io/worker:REPLACE
+          image: gcr.io/zeebe-io/worker:{{ imageTag }}
           imagePullPolicy: Always
           env:
             - name: JAVA_OPTIONS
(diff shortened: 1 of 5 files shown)


go-chaos/internal/manifests/worker.yaml
```diff
@@ -16,15 +16,15 @@ spec:
     spec:
       containers:
         - name: worker
-          image: gcr.io/zeebe-io/worker:REPLACE
+          image: gcr.io/zeebe-io/worker:{{ imageTag }}
```
npepinpe, line 19: :thought_balloon: Because we parameterize two things now, I figured we can start using a more explicit notation than just `REPLACE`. If we parameterize 3 or more, we should look into using go template ;)

npepinpe: Rule of three :smile:

Issue #458, opened 2023-12-11
zbchaos supports broker scaling


oleschoenburg:
Similar to the `backup` command, `zbchaos` should be able to use the `/cluster` actuator to request the current topology and issue a scale request.

oleschoenburg:
I'll close this, we'll add additional functionality if/when needed

PR #459, opened 2023-12-11
feat(zbchaos): query current cluster topology

oleschoenburg:
Adds a new command `zbchaos cluster status` that uses  `GET /actuator/cluster` under the hood and pretty-prints the deserialized response.

Part 1 of X for https://github.com/zeebe-io/zeebe-chaos/issues/458

### DIFF
--- a/go-chaos/cmd/cluster.go
+++ b/go-chaos/cmd/cluster.go
@@ -99,26 +99,52 @@ func waitForChange(flags *Flags) error {
 		if err != nil {
 			return err
 		}
-		if topology.LastChange != nil && topology.LastChange.Id == int64(flags.changeId) {
-			if topology.LastChange.Status == "COMPLETED" {
-				internal.LogInfo("Change %d completed successfully", flags.changeId)
-				return nil
-			} else {
-				return fmt.Errorf("change %d failed with status %s", flags.changeId, topology.LastChange.Status)
-			}
-		} else if topology.LastChange != nil && topology.LastChange.Id > int64(flags.changeId) {
+		changeStatus := describeChangeStatus(topology, int64(flags.changeId))
+		switch changeStatus {
+		case ChangeStatusCompleted:
+			internal.LogInfo("Change %d completed successfully", flags.changeId)
+			return nil
+		case ChangeStatusFailed:
+			internal.LogInfo("Change %d failed with status %s", flags.changeId, topology.LastChange.Status)
+			return fmt.Errorf("change %d failed with status %s", flags.changeId, topology.LastChange.Status)
+		case ChangeStatusOutdated:
 			internal.LogInfo("Change %d is outdated but was most likely completed successfully, latest change is %d", flags.changeId, topology.LastChange.Id)
 			return nil
-		} else if topology.PendingChange != nil && topology.PendingChange.Id == int64(flags.changeId) {
+		case ChangeStatusPending:
 			competed := len(topology.PendingChange.Completed)
 			pending := len(topology.PendingChange.Pending)
 			total := competed + pending
 			internal.LogInfo("Change %d is %s with %d/%d operations complete", flags.changeId, topology.PendingChange.Status, competed, total)
-		} else {
+		case ChangeStatusUnknown:
 			internal.LogInfo("Change %d not yet started", flags.changeId)
 		}
-
 		time.Sleep(5 * time.Second)
+	}
+}
+
+type ChangeStatus string
+
+const (
+	ChangeStatusOutdated  ChangeStatus = "OUTDATED"
+	ChangeStatusCompleted ChangeStatus = "COMPLETED"
+	ChangeStatusFailed    ChangeStatus = "FAILED"
+	ChangeStatusPending   ChangeStatus = "PENDING"
+	ChangeStatusUnknown   ChangeStatus = "UNKNOWN"
+)
+
+func describeChangeStatus(topology *CurrentTopology, changeId int64) ChangeStatus {
+	if topology.LastChange != nil && topology.LastChange.Id == changeId {
+		if topology.LastChange.Status == "COMPLETED" {
+			return ChangeStatusCompleted
+		} else {
+			return ChangeStatusFailed
+		}
+	} else if topology.LastChange != nil && topology.LastChange.Id > changeId {
+		return ChangeStatusOutdated
+	} else if topology.PendingChange != nil && topology.PendingChange.Id == changeId {
+		return ChangeStatusPending
+	} else {
+		return ChangeStatusUnknown
 	}
 }
 
--- a/go-chaos/cmd/cluster.go
+++ b/go-chaos/cmd/cluster.go
@@ -94,7 +94,10 @@ func waitForChange(flags *Flags) error {
 	closePortForward := k8Client.MustGatewayPortForward(port, port)
 	defer closePortForward()
 
-	for {
+	interval := time.Second * 5
+	timeout := (time.Minute * 25)
+	iterations := int(timeout / interval)
+	for i := 0; i < int(iterations); i++ {
 		topology, err := queryTopology(port)
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+		Short: "Waits for a topology change to complete",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return waitForChange(flags)
+		},
+	}
+
+	rootCmd.AddCommand(clusterCommand)
+	clusterCommand.AddCommand(statusCommand)
+	clusterCommand.AddCommand(waitCommand)
+	waitCommand.Flags().IntVar(&flags.changeId, "changeId", 0, "The id of the change to wait for")
```
npepinpe, line 52: :upside_down_face: I know almost all Go linters complain about using `Id`, and always want to change to `ID`. While I have no strong feelings about it, if we ever add them I imagine this will happen :smile: 

I guess that applies to all the `Id` you have below. :shrug: 

```suggestion
	waitCommand.Flags().IntVar(&flags.changeID, "changeId", 0, "The id of the change to wait for")
```

go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+
+	port := 9600
+	closePortForward := k8Client.MustGatewayPortForward(port, port)
+	defer closePortForward()
+
+	topology, err := queryTopology(port)
+	if err != nil {
+		return err
+	}
+	formatted, err := json.MarshalIndent(topology, "", "  ")
```
npepinpe, line 74: :exclamation: Cool, wasn't aware of that

go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+		} else if topology.PendingChange != nil && topology.PendingChange.Id == int64(flags.changeId) {
+			competed := len(topology.PendingChange.Completed)
+			pending := len(topology.PendingChange.Pending)
+			total := competed + pending
+			internal.LogInfo("Change %d is %s with %d/%d operations complete", flags.changeId, topology.PendingChange.Status, competed, total)
+		} else {
+			internal.LogInfo("Change %d not yet started", flags.changeId)
+		}
+
+		time.Sleep(5 * time.Second)
```
npepinpe, line 121: :wrench: I would add some maximum time out or something, this looks to me like it could potentially loop forever? I know we assume the change will eventually complete, but still... :sweat_smile:

npepinpe (approved):
LGTM, though I haven't tested it yet. I'll try it out tomorrow (though I assume you already tested it ;))

go-chaos/cmd/cluster.go
Zelldon: Yes we should definitely fail at some point.

oleschoenburg: Yeah, good point :+1: Picking a good value is tricky though, I'd expect scaling to take anywhere from 5 minutes to >30 minutes..

Using that in an automated chaos test is problematic anyway because we'll run into job timeouts... :thinking:

go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+			return nil
+		} else if topology.PendingChange != nil && topology.PendingChange.Id == int64(flags.changeId) {
+			competed := len(topology.PendingChange.Completed)
+			pending := len(topology.PendingChange.Pending)
+			total := competed + pending
+			internal.LogInfo("Change %d is %s with %d/%d operations complete", flags.changeId, topology.PendingChange.Status, competed, total)
+		} else {
+			internal.LogInfo("Change %d not yet started", flags.changeId)
+		}
+
```
Zelldon, line 120: Maybe you can extract this into a separate function. Would like to see an unit test, where we give an topology in and you get an result out, like valid or invalid, etc.

go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func queryTopology(port int) (*CurrentTopology, error) {
```
Zelldon, line 125: 🔧 Would be great if you could add a IT for it

go-chaos/cmd/cluster.go
```diff
@@ -0,0 +1,190 @@
...
+	var topology CurrentTopology
+	err = json.Unmarshal(body, &topology)
+	if err != nil {
+		return nil, err
+	}
+	return &topology, nil
+}
+
+type CurrentTopology struct {
+	Version       int32
```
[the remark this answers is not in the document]
Zelldon: Does this work without the `json` annotations?

Zelldon (commented):
Thanks @oleschoenburg 👍🏼 I had some comments see below :)

go-chaos/cmd/cluster.go
oleschoenburg: So far I haven't seen a linter complain about `bpmnProcessId`, `partitionId` etc. I really dislike `changeID` so I think I'll stick with `changeId` for now

go-chaos/cmd/cluster.go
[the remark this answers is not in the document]
oleschoenburg: Yes, it works by default

go-chaos/cmd/cluster.go
Zelldon: W00t 30 minutes 😆 😅

go-chaos/cmd/cluster.go
oleschoenburg: I've added a minimal one now. None of the other commands have integration tests so I hope I did it in a way that is acceptable: https://github.com/zeebe-io/zeebe-chaos/pull/459/commits/ffd0a20c687e7616b605de8be0edf687e0c5a8f1

go-chaos/cmd/cluster.go
oleschoenburg: I've set a timeout of 25 minutes now

go-chaos/cmd/cluster.go
npepinpe: No integration tests for other commands you say? 

shock

go-chaos/integration/cluster_cmd_integration_test.go
```diff
@@ -0,0 +1,70 @@
...
+	require.NotNil(t, topology.Brokers)
+	require.Equal(t, 1, len(topology.Brokers))
+	require.Equal(t, int32(1), topology.Version)
+	require.Nil(t, topology.LastChange)
+	require.Nil(t, topology.PendingChange)
+}
+
+func CreateZeebeContainer(t *testing.T, ctx context.Context) testcontainers.Container {
+	req := testcontainers.ContainerRequest{
+		Image:        "camunda/zeebe:SNAPSHOT",
```
Zelldon, line 60: 🔧 Lets pin it to a version where we know this has it. Or is it not released yet?

go-chaos/integration/cluster_cmd_integration_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// then
+	require.NoError(t, err)
+	require.NotNil(t, topology)
+	require.NotNil(t, topology.Brokers)
+	require.Equal(t, 1, len(topology.Brokers))
+	require.Equal(t, int32(1), topology.Version)
+	require.Nil(t, topology.LastChange)
+	require.Nil(t, topology.PendingChange)
+}
```
Zelldon, line 56: Yeah I guess this doesn't help much, so we will see only if this works if you run an experiment for it.

Zelldon (approved):
Thanks @oleschoenburg 👍🏼

go-chaos/integration/cluster_cmd_integration_test.go
oleschoenburg: I can pin it to 8.4.0-alpha2 but that way we won't notice when zbchaos stops working with newer Zeebe versions, for example if we change the API

go-chaos/cmd/cluster.go
```diff
@@ -99,29 +99,55 @@ func waitForChange(flags *Flags) error {
...
+
+const (
+	ChangeStatusOutdated  ChangeStatus = "OUTDATED"
+	ChangeStatusCompleted ChangeStatus = "COMPLETED"
+	ChangeStatusFailed    ChangeStatus = "FAILED"
+	ChangeStatusPending   ChangeStatus = "PENDING"
+	ChangeStatusUnknown   ChangeStatus = "UNKNOWN"
+)
+
+func describeChangeStatus(topology *CurrentTopology, changeId int64) ChangeStatus {
```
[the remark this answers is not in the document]
oleschoenburg: Our API doesn't offer an endpoint to query "what's the status of change X". If we would offer that, the endpoint would return a similar enum as we have defined here, yeah.

PR #461, opened 2023-12-13
fix: use correct JDK_JAVA_OPTIONS

Zelldon:
We need to use the newer java options, in order to configure our application otherwise it will not be picked up.


npepinpe (approved):
:rocket:

Zelldon:
For posterity https://github.com/zeebe-io/benchmark-helm/pull/20

PR #464, opened 2023-12-14
feat(zbchaos): add command for broker scaling

oleschoenburg:
Adds a new command that makes an API call to Zeebe to scale to the given amount of brokers and then scales up the statefulset accordingly. The command knows if it is scaling up or down to scale the statefulset first or last. The command always waits for completion as it can only scale down the statefulset after the change is applied.

Usage:

```
$ zbchaos cluster scale --brokers 5
```

Part 2/X of #458

### DIFF
--- a/go-chaos/cmd/cluster.go
+++ b/go-chaos/cmd/cluster.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
+
+	changeResponse, err := requestBrokerScaling(port, flags.brokers)
+	if err != nil {
+		return err
+	}
+	formatted, err := json.MarshalIndent(changeResponse, "", "  ")
+	if err != nil {
+		return err
+	}
+	fmt.Println(string(formatted))
```
Zelldon, line 88: ❌ Sorry I have overseen this before. Can you change this and also on the other subcommand to use the `internal.LogInfo` or `internal.LogVerbose` depending on the verbose flag you can then control whether it is logged or not. Furthermore, it adds additional context with json logging enabled.

go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
+
+	return nil
+}
+
+func requestBrokerScaling(port int, brokers int) (*ChangeResponse, error) {
+	brokerIds := make([]int32, brokers)
+	for i := 0; i < brokers; i++ {
+		brokerIds[i] = int32(i)
+	}
+	url := fmt.Sprintf("http://localhost:%d/actuator/cluster/brokers", port)
```
Zelldon, line 103: ❓ So I just give it a list of brokers Ids to scale? 🤔 What was the idea behind this? Could I also give a list which starts at 10 to 15? Why not having an endpoint where I can do like `inc` by one. Out of curiosity.

go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
 	waitCommand.Flags().IntVar(&flags.changeId, "changeId", 0, "The id of the change to wait for")
+	clusterCommand.AddCommand(scaleCommand)
+	scaleCommand.Flags().IntVar(&flags.brokers, "brokers", 0, "The amount of brokers to scale to")
+}
+
+func scaleCluster(flags *Flags) error {
+	k8Client, err := createK8ClientWithFlags(flags)
+	if err != nil {
+		panic(err)
+	}
```
Zelldon, line 69:
```suggestion
	ensureNoError(err)
```

go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
+
+	err = k8Client.AwaitReadiness()
+	if err != nil {
+		return err
+	}
```
Zelldon, line 74: A reason to not panic here? If not you can use.
```suggestion
  ensureNoError(err)
```

go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
+	resp, err := http.Post(url, "application/json", bytes.NewReader(request))
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
+		return nil, fmt.Errorf("scaling failed with code %d", resp.StatusCode)
+	}
+
+	defer func(Body io.ReadCloser) {
+		_ = Body.Close()
```
Zelldon, line 117: ❓ Wondering whether you want to close this even if the body was nil, or faulty and not readable. Might make sense to move the defer block 🤔

go-chaos/cmd/cluster.go
```diff
@@ -45,11 +46,88 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
...
+			return scaleCluster(flags)
+		},
+	}
 
 	rootCmd.AddCommand(clusterCommand)
 	clusterCommand.AddCommand(statusCommand)
 	clusterCommand.AddCommand(waitCommand)
 	waitCommand.Flags().IntVar(&flags.changeId, "changeId", 0, "The id of the change to wait for")
+	clusterCommand.AddCommand(scaleCommand)
+	scaleCommand.Flags().IntVar(&flags.brokers, "brokers", 0, "The amount of brokers to scale to")
```
Zelldon, line 62: ❓ Is zero really the default you want to have here? I would recommend it is either three ( our standard setting) or one, or something you would normally use in experiments.

Zelldon (changes requested):
Great @oleschoenburg 👍🏼 

One remark regarding the printing, others are more nitpicks, questions/ideas.

go-chaos/cmd/cluster.go
oleschoenburg: Oh yeah good point :+1:

go-chaos/cmd/cluster.go
oleschoenburg: Yes, that's correct. You can give it ids 10-15 if you want :+1: 

An inc by one endpoint could be added to the API. So far only have two endpoints for this, adding a broker with a given id and updating the list of brokers completely (this is what we use here)

oleschoenburg:
@Zelldon Thanks for the review so far :+1: I just realized that this doesn't work for scaling down because there we need to make the API request first, wait for the change to complete and only then scale down the statefulset. I'll fix this first before merging.

go-chaos/cmd/cluster.go
oleschoenburg: "The http Client and Transport guarantee that Body is always non-nil"
I think it's fine to always call close? :shrug:

go-chaos/cmd/cluster.go
oleschoenburg: It's a required flag though so the default of 0 is unused anyway. I'd leave it as is/

oleschoenburg:
Rebased on main to get the local port improvement

Zelldon (approved):
🚀 Cool stuff! Thanks @oleschoenburg

PR #465, opened 2023-12-18
feat(zbchaos): port-forward to random local port

oleschoenburg:
All commands now provide `0` as the local port which results in a free port picked at random. This makes it easier to run two zbchaos commands at the same time without getting port conflicts.


Zelldon (approved):
❤️ Thanks @oleschoenburg 👍🏼

PR #473, opened 2023-12-29
experiment: add experiment for borker-scaling

deepthidevaki:
Add an experiment to verify that brokers can be scaled up and the processing continues after scaling. To verify processing continues, a message is published to partition 3. After scaling we expect it to be correlated to a new instance. 

The experiment can only be run on Production - S because it assumes a cluster size of 3 and partitions 3. 

I have not tested it. But since we have already manually tested `zbchaos scale` command, I assume this works.

related to https://github.com/camunda/zeebe/issues/15752


Zelldon:
This is part of v1.6.0 https://github.com/zeebe-io/zeebe-chaos/releases/tag/zbchaos-v1.6.0

PR #490, opened 2024-02-06
fix(zbchaos): retry SaaS detection and panic if it fails instead of falling back to Self-Managed mode

### LINKED ISSUE #488: Detection of SaaS vs SM environment incorrectly fallback to SM
Observed the following error when running chaos tests against our SaaS cluster
```
Failed to retrieve SaaS CRD, fallback to self-managed mode. Get "https://34.79.82.85/apis/cloud.camunda.io/v1alpha1/zeebeclusters/e8196081-1341-4b05-a7fb-b3c7dcf8f646": dial tcp 34.79.82.85:443: i/o timeout"
```

As a result "workers" for a test were deployed using SM configuration which uses a different service name for the gateways. The test failed because the workers cannot talk to the Zeebe gateway because of the incorrect service neame.

** Expected behavior **
 
The environment is detected correctly. 
Timeouts should not considered as a valid response to determine the environment. It should be retried.

oleschoenburg:
I have added a dependency on a generic exponential backoff library because I didn't feel like writing this on my own. If querying for the CR still fails after retries are exhausted, we panic instead of just falling back to Self-Managed mode. I think this is preferable over retrying forever.

Closes #488

### DIFF
--- a/go-chaos/internal/saas.go
+++ b/go-chaos/internal/saas.go
@@ -19,7 +19,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/cenkalti/backoff/v4"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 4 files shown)


oleschoenburg:
Unfortunately this breaks test that depend on the fallback behavior. For example `Test_CreateK8ClientWithPath` which creates a client that tries to connect to https://127.0.0.1:46807. When building the client, we test if it's SaaS or SM. Since the URL is not reachable we can't query the CR and thus can't determine if it's SaaS or not.

Seems to me that we can't fix https://github.com/zeebe-io/zeebe-chaos/issues/488 without breaking a lot of tests :sweat_smile:

go-chaos/internal/saas.go
```diff
@@ -59,11 +61,13 @@ func (c K8Client) isSaaSEnvironment() bool {
...
-	resource, err := c.DynamicClient.Resource(zeebeCrd).Get(context.TODO(), clusterId, meta.GetOptions{})
+	resource, err := backoff.RetryWithData(func() (*unstructured.Unstructured, error) {
+		return c.DynamicClient.Resource(zeebeCrd).Get(context.TODO(), clusterId, meta.GetOptions{})
+	}, backoff.NewExponentialBackOff())
 
 	if err != nil {
-		LogInfo("Failed to retrieve SaaS CRD, fallback to self-managed mode. %v", err)
-		return false
+		LogError("Failed to check for SaaS CRD, can't proceed without knowing if this is SaaS or Self-Managed: %v", err)
+		panic(err)
```
Zelldon, line 70: Let's not panic here :) If you want we can take a look at it together tomorrow.

oleschoenburg: But then what? We don't know if it's SaaS or not

Zelldon: My point I wanted to make is the following:

I don't want to have the panics in the internal package, you can see it as the lib or backend. It would be same as a library in java just does exit(1), it is unexpected and not nice. 

I would like to see to return an error and handle this by the consumer of the method. 

I think we have here three cases to distinguish: a) CRD exists - SaaS b) CRD not exist - SM c) error -> return error. If currently not existing also causes an error (we have to find out/check this I guess, then we need to somehow need to distinguish these error cases.

oleschoenburg: Okay that part makes sense to me but I'm not sure how to handle the errors. We check for SaaS when building the k8sclient so very early on for every request. What can we do if that fails?

PR #493, opened 2024-02-09
Saas detection error handling

### LINKED ISSUE #488: Detection of SaaS vs SM environment incorrectly fallback to SM

Zelldon:
Paired with @oleschoenburg on this:

 * We agreed on to "simply" pass the error to the front, to the respective commands so they can either fail or do what ever
 * For errors indicating resource not-found we decided to return false, and use SM setting (as we expect this CRD not in the SM env)
 * We wrote some more tests for the error handling
 * I had to refactor a bit the client creation, so we test the right things here

closes #488 
closes #490

### DIFF
--- a/go-chaos/internal/k8helper.go
+++ b/go-chaos/internal/k8helper.go
@@ -49,6 +49,26 @@ func CreateK8Client(kubeConfigPath string, namespace string) (K8Client, error) {
 }
 
 func createK8Client(settings KubernetesSettings) (K8Client, error) {
+	client, err := internalCreateClient(settings)
+	if err != nil {
+		return client, err
+	}
+
+	client.SaaSEnv, err = client.isSaaSEnvironment()
+	if err != nil {
+		return client, err
+	}
+
+	if client.SaaSEnv {
+		LogVerbose("Running experiment in SaaS environment.")
+	} else {
+		LogVerbose("Running experiment in self-managed environment.")
+	}
+
+	return client, nil
+}
+
+func internalCreateClient(settings KubernetesSettings) (K8Client, error) {
 	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		&clientcmd.ClientConfigLoadingRules{ExplicitPath: settings.kubeConfigPath},
 		&clientcmd.ConfigOverrides{Context: api.Context{Namespace: settings.namespace}})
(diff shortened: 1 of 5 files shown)


go-chaos/internal/k8helper.go
```diff
@@ -49,6 +49,26 @@ func CreateK8Client(kubeConfigPath string, namespace string) (K8Client, error) {
...
 
 func createK8Client(settings KubernetesSettings) (K8Client, error) {
+	client, err := internalCreateClient(settings)
+	if err != nil {
+		return client, err
+	}
+
+	client.SaaSEnv, err = client.isSaaSEnvironment()
+	if err != nil {
+		return client, err
```
oleschoenburg, line 59: :upside_down_face: Super nitpicky but I think we are not supposed to return the client in this case? 

```suggestion
		return nil, err
```

Zelldon: Makes sense thanks 👍🏼

Zelldon:
```
go build -o zbchaos main.go
# github.com/zeebe-io/zeebe-chaos/go-chaos/internal
Error: internal/k8helper.go:5[9](https://github.com/zeebe-io/zeebe-chaos/actions/runs/7843937019/job/21405291067#step:7:10):[10](https://github.com/zeebe-io/zeebe-chaos/actions/runs/7843937019/job/21405291067#step:7:11): cannot use nil as K8Client value in return statement
make: *** [Makefile:11: build] Error 1
Error: Process completed with exit code 2.
```

😿

Zelldon:
@oleschoenburg I would remove the last commit, we would either have to return K8Client{} or simply the client obj we had before. wdyt?

oleschoenburg:
Ah how annoying. Yeah, just drop the last commit.

PR #508, opened 2024-03-11
fix: update image registry to correct one

berkaycanbc:
**Description**

Based on the [Slack thread](https://camunda.slack.com/archives/CQZGZ882Z/p1710145484340359?thread_ts=1710144915.404369&cid=CQZGZ882Z) our image registry is changed for int and dev. That causes our chaos tests to fail. This PR is attempting to fix the registry.


nicpuppa (approved):
LGTM 🚀

PR #509, opened 2024-03-11
Add commands for force failover to a region

deepthidevaki:
Related to https://github.com/camunda/zeebe-e2e-test/issues/421 

This PR adds commands for 
1. Force failover to one region `zbchaos forceFailover --regions=2 --regionId=1`
    * Removes region 0 and keeps the brokers from region 1 
2.  And adding the region back `zbchaos scale --brokers=4 --replicationFactor=4`
    * Re-uses the scale command, and added the parameter `replicationFactor`

The commands are manually tested on a benchmark cluster.

Also adds an optional `awaitReadiness` flag to `dataloss recover` command. When using the new API for failover, all brokers in the failed region must be started in parallel. Waiting for one broker to be ready before the others are started can prevent the failback from completing.

### DIFF
--- a/go-chaos/cmd/cluster.go
+++ b/go-chaos/cmd/cluster.go
@@ -53,15 +53,27 @@ func AddClusterCommands(rootCmd *cobra.Command, flags *Flags) {
 			return scaleCluster(flags)
 		},
 	}
+	var forceFailoverCommand = &cobra.Command{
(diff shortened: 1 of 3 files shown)


npepinpe (approved):
Tentatively approve, I think we need to sync before I fully understand, since I missed adding the API endpoint and so on

PR #531, opened 2024-04-30
Remove security enforcement label from namespace in SaaS

Zelldon:
In SaaS we recently introduced a label that indicates to prevent certain actions (security related), see related [slack thread ](https://camunda.slack.com/archives/CT702EPFH/p1714475896279759)

In order to make sure that our experiments, and actions are successful in SaaS
we need to make sure that reconciliation is paused and the
security enforcement label is removed from the corresponding
target namespace. That is now always done when creating a client for a SaaS environment.

After doing so we can get further privileges, that are needed for
actions like putting stress on the CPU, network partition, etc.


oleschoenburg (approved):
Thanks! :rocket:

PR #617, opened 2024-11-18
feat: migrate zeebe's backup api from /backups to /backup-runtime

### LINKED ISSUE #616: Migrate zeebe's backup api from `/actuator/bacups` to `/actuator/backup-runtime`
See https://github.com/camunda/camunda/issues/24456 for reference

entangled90:
Updates the actuator endpoints to match https://github.com/camunda/camunda/issues/24456

closes #616

### DIFF
--- a/go-chaos/cmd/backup.go
+++ b/go-chaos/cmd/backup.go
@@ -195,7 +195,7 @@ func takeBackup(flags *Flags) error {
 
 	port, closePortForward := k8Client.MustGatewayPortForward(0, 9600)
 	defer closePortForward()
-	url := fmt.Sprintf("http://localhost:%d/actuator/backups/%s", port, flags.backupId)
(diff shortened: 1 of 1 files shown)


lenaschoenburg (approved):
Thanks!

entangled90:
Should I change the path depending on the version number?

lenaschoenburg:
As far as I know we have no automated tests so nothing should break. Detecting the right version would be fancy but I don't think it's strictly required.

entangled90:
I see that all other opened PR fail with the same error, should I proceed with the merge?

lenaschoenburg:
I'd say yes unless @Zelldon vetos.

Zelldon:
> As far as I know we have no automated tests so nothing should break. Detecting the right version would be fancy but I don't think it's strictly required. 

Wouldnt the experiment itself break then?

lenaschoenburg:
> Wouldnt the experiment itself break then?

Which experiment? I found none that made use of backups.

PR #654, opened 2025-07-02
Mob session

panagiotisgts:
- Allow chaos client to retrieve k8s pod Labels/Names from the new deployment labels
- Adjust tests to align with updated k8s client dependency

### DIFF
--- a/go-chaos/internal/labels.go
+++ b/go-chaos/internal/labels.go
@@ -16,6 +16,7 @@ package internal
 
 import (
 	"context"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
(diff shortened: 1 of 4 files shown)


entangled90:
@ChrisKujawa should we merge this anyway since main is already broken or is the plan to fix all tests in this PR?

go-chaos/internal/labels.go
```diff
@@ -79,6 +80,30 @@ func (c K8Client) getGatewayLabels() string {
...
+}
+
+func getSaasCoreLabels() string {
+	// For backwards compatability the brokers kept the core labels, for a statefulset the labels are not modifiable
+	// To still be able to distinguish the standalone core with the broker, the gateway got a new label.
+	labelSelector := metav1.LabelSelector{
+		MatchLabels: map[string]string{"app.kubernetes.io/app": "zeebe-gateway", "app.kubernetes.io/component": "standalone-gateway"},
+	}
+	return labels.Set(labelSelector.MatchLabels).String()
+}
```
ChrisKujawa, line 97: 🤔 ❓ I think this is not necessary, as we don't have core labels in SaaS

go-chaos/internal/statefulset_test.go
```diff
@@ -72,10 +72,13 @@ func Test_ShouldReturnErrorForNonExistingStatefulSetInSaaS(t *testing.T) {
...
-	statefulset, err := k8Client.GetZeebeStatefulSet()
+	_, err := k8Client.GetZeebeStatefulSet()
 
 	// then
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "statefulsets.apps \"zeebe\" not found")
-	assert.Nil(t, statefulset)
+
+	// API No longer returns nil pointer but empty initialized struct
+	//assert.Nil(t, statefulset)
```
ChrisKujawa, line 82: ❓ Do we want to remove this?

go-chaos/internal/labels.go
entangled90: Good point, I will push the check further upwards

ChrisKujawa (approved):
Thanks 👍🏼 

one thing but not a blocker :)

PR #662, opened 2025-07-09
Add client credentials to worker

entangled90:
Credentials are fetched from the job variables and added to the flags of the `zbchaos` process created in `HandleZbChaosJob`.

A real json taken from test bench has been used to verify correct parsing from json

`ContactPoint` is necessary?

### DIFF
--- a/go-chaos/worker/chaos_worker_test.go
+++ b/go-chaos/worker/chaos_worker_test.go
@@ -285,6 +285,19 @@ func Test_shouldUnmarshalAuthCredentials(t *testing.T) {
 	assert.Equal(t, authentication.ClientSecret, "superSecret")
 }
 
+func Test_shouldCreateFlagsCorrectly(t *testing.T) {
+	auth := createZbChaosVariables().AuthenticationDetails
+
+	flags := auth.toFlags()
+	expectedFlags := []string{
+		"--authServer=https://auth.com/url",
+		"--audience=zeebe.com",
+		"--clientId=randomClientId",
+		"--clientSecret=superSecret",
+	}
+	assert.Equal(t, expectedFlags, flags)
+}
+
 func createVariablesAsJson() (string, error) {
 	variables := createZbChaosVariables()
 


go-chaos/worker/chaos_worker.go
```diff
@@ -40,12 +40,31 @@ type ChaosProvider struct {
 
 type AuthenticationProvider struct {
 	Audience         string
-	AuthorizationUrl string
+	AuthorizationURL string
 	ClientId         string
 	ClientSecret     string
 	ContactPoint     string
```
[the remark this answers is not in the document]
entangled90: @ChrisKujawa what do you prefer?

go-chaos/worker/chaos_worker.go
```diff
@@ -40,10 +40,29 @@ type ChaosProvider struct {
 
 type AuthenticationProvider struct {
 	Audience         string
-	AuthorizationUrl string
+	AuthorizationURL string
 	ClientId         string
 	ClientSecret     string
-	ContactPoint     string
+	// ContactPoint     string
```
ChrisKujawa, line 46: 👍🏼 We can remove it completelty - we dont need it as we port-forward and directly connect to the gatway/broker.

go-chaos/worker/chaos_worker.go
```diff
@@ -40,10 +40,29 @@ type ChaosProvider struct {
...
+	}
+	if c.ClientId != "" {
+		flags = append(flags, fmt.Sprintf("--clientId=%s", c.ClientId))
+	}
+	if c.ClientSecret != "" {
+		flags = append(flags, fmt.Sprintf("--clientSecret=%s", c.ClientSecret))
+	}
+
+	return flags
 }
```
ChrisKujawa, line 66: 👍🏼 Nice

go-chaos/worker/chaos_worker.go
```diff
@@ -197,7 +221,7 @@ func getTargetClusterVersion(namespace string) string {
 	port, closeFn := k8Client.MustGatewayPortForward(0, 26500)
 	defer closeFn()
 
-	zbClient, err := internal.CreateZeebeClient(port)
+	zbClient, err := internal.CreateZeebeClient(port, credentials)
```
ChrisKujawa, line 224: ❓ 🔧 We could also pass in here nil and then remove the parameter of the `getTargetClusterVersion` function

go-chaos/worker/chaos_worker_test.go
```diff
@@ -121,14 +121,19 @@ func Test_ShouldHandleCommand(t *testing.T) {
 	// then
 	assert.True(t, fakeJobClient.Succeeded)
 	assert.Equal(t, 123, fakeJobClient.Key)
-	var expectedArgs = []string{
+	expectedArgs := []string{
 		"--namespace", "clusterId-zeebe",
+		"--authServer=https://auth.com/url",
+		"--audience=zeebe.com",
+		"--clientId=randomClientId",
+		"--clientSecret=superSecret",
```
ChrisKujawa, line 129: ❓ We need to add the values also to the job variables right? 🤔

go-chaos/worker/chaos_worker_test.go
```diff
@@ -241,17 +251,40 @@ func Test_ShouldFailJobWhenHandleFails(t *testing.T) {
...
 		"--verbose",
 		"--jsonLogging",
 		"--dockerImageTag",
-		"test"}
+		"test",
+	}
 	assert.Equal(t, expectedArgs, appliedArgs)
 }
 
+func Test_shouldUnmarshalAuthCredentials(t *testing.T) {
```
ChrisKujawa, line 270: 🔧 Lets add also a test for the flags conversation

go-chaos/internal/manifests/worker.yaml
```diff
@@ -35,4 +35,4 @@ spec:
               memory: 2Gi
             requests:
               cpu: 1
-              memory: 512Mi
\ No newline at end of file
+              memory: 512Mi
```
ChrisKujawa, line 38: ❌ Actually this worker deployment also need the credentials otherwise some experiments will not work, like the ones where we depend on the workers (where we await the instance completion)

ChrisKujawa (approved):
👍🏼 🚀 Thanks @entangled90 good work!

There is one missing piece, we have to adjust the benchmark worker manifest as well. But we can also do this in a separate PR.

go-chaos/cmd/root.go
```diff
@@ -148,3 +161,26 @@ func Execute() {
...
+	}
+	if len(flags.audience) == 0 {
+		return nil
+	}
+	if len(flags.clientSecret) == 0 {
+		return nil
+	}
+	if len(flags.clientId) == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
entangled90: Thanks, that's exactly what we needed. 
I assume all of this must be passed right?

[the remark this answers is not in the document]
ChrisKujawa: Yes exactly

go-chaos/internal/manifests/worker.yaml
entangled90: Go templating is used to replace all variables

PR #685, opened 2025-09-09
fix: failover must remove brokers from other region

### LINKED ISSUE #684: Multi-region force failover job is not removing the correct region
Context: [E2E Multi-region test failure](https://camunda.slack.com/archives/C013MEVQ4M9/p1756717676941539) 

First we see request to remove broker 0 and 2
```
Requesting scaling http://localhost:46373/actuator/cluster?force=true with input {"brokers":{"remove":[2,0]},"partitions":null} 
```
Then we see the disk of brokers 1 and 3 is deleted.

```
Deleting PVC data-zeebe-1 in namespace 22be1922-f487-4a66-98c7-3ed46d5b9c5b-zeebe 
Deleting PVC data-zeebe-3 in namespace 22be1922-f487-4a66-98c7-3ed46d5b9c5b-zeebe 
```

This is not what we want to do in this test. To simulate failure of one region, we want to remove the brokers first via the api and then restart them after deleting the data. But here the data of the surviving brokers have been deleted, thus ending up in a full data loss situation.

This is probably introduced due to recent changes in the cluster scale api.

Expected behaviour
* When removing region 1, it should first shutdown pods zeebe-0 and zeebe-2. Then remove disk data-zeebe-0 and data-zeebe-2
* When removing region 1, it should remove pods and data for zeebe-1 and zeebe-3

lenaschoenburg:
Fixes a bug introduced with https://github.com/camunda/zeebe-chaos/pull/682 where we were removing brokers from the region that we wanted to failover _to_. We must remove the brokers from the _other_ region instead.

Example of bad behavior before:

```
Running command with args: [--cluster forceFailover --regions=2 --regionId=0] 
Requesting scaling http://localhost:41617/actuator/cluster?force=true with input {"brokers":{"remove":[2,0]},"partitions":null}
```

Here, brokers 0 and 2 are from region 0, we want to remove brokers 1 and 3 instead.

Closes #684

### DIFF
--- a/go-chaos/cmd/cluster.go
+++ b/go-chaos/cmd/cluster.go
@@ -357,9 +357,9 @@ func forceFailover(flags *Flags) error {
 		return fmt.Errorf("cluster is already scaling")
 	}
 
-	brokersInRegion := getBrokers(currentTopology, flags.regions, flags.regionId)
+	brokersToRemove := getBrokersToRemove(currentTopology, flags.regions, flags.regionId)
 
-	changeResponse, err := sendScaleRequest(port, brokersInRegion, 0, true, -1)
+	changeResponse, err := sendScaleRequest(port, brokersToRemove, 0, true, -1)
 	ensureNoError(err)
 
 	timeout := time.Minute * 5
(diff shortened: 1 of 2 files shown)


go-chaos/cmd/cluster.go
```diff
@@ -369,10 +369,10 @@ func forceFailover(flags *Flags) error {
 	return nil
 }
 
-func getBrokers(topology *CurrentTopology, regions int32, regionId int32) []int32 {
+func getBrokersToRemove(topology *CurrentTopology, regions int32, regionId int32) []int32 {
 	brokersInRegion := make([]int32, 0)
```
deepthidevaki, line 373: :wrench: May be rename to `brokerInOtherRegion`

deepthidevaki (approved):
Thanks

PR #686, opened 2025-09-18
fix: gateway are selected via labels on the service

entangled90:
With the new deployment configuration for self managed, gateways are not deployed anymore as they are part of the pods named `camunda-{n}`.

With this change, when port forwading, a random pod is selected from the gateway service with the specified port.

### DIFF
--- a/go-chaos/internal/helper_test.go
+++ b/go-chaos/internal/helper_test.go
@@ -144,3 +144,16 @@ func (c *K8Client) createSaaSNamespace(t *testing.T) {
 	_, err := c.Clientset.CoreV1().Namespaces().Create(context.TODO(), &namespace, metav1.CreateOptions{})
 	require.NoError(t, err)
 }
+
+// CreateServiceWithSelector creates a ClusterIP service with the given name, selector labels, and ports.
+func (c K8Client) CreateServiceWithSelector(t *testing.T, name string, selector map[string]string, ports []v1.ServicePort) {
+	if len(ports) == 0 {
+		ports = []v1.ServicePort{{Port: 26500}}
+	}
+	svc := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Labels: selector},
+		Spec:       v1.ServiceSpec{Selector: selector, Ports: ports, Type: v1.ServiceTypeClusterIP},
+	}
+	_, err := c.Clientset.CoreV1().Services(c.GetCurrentNamespace()).Create(context.TODO(), svc, metav1.CreateOptions{})
+	require.NoError(t, err)
+}
--- a/go-chaos/internal/pods.go
+++ b/go-chaos/internal/pods.go
@@ -176,7 +176,7 @@ func (c K8Client) AwaitReadinessWithTimeout(timeout time.Duration, tickTime time
 	for {
 		select {
 		case <-timedOut:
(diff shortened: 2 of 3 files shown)


entangled90:
I can test again today to be sure, but in general we select the service by label `zeebe-gateway` and not by name, so I hope it should work :wink:

ChrisKujawa:
> Tested again and it works with a fresh deployed cluster 

Be aware that the changes in the platform chart are not in the benchmark chart as we use pinned versions

entangled90:
PR #687 was merged in this branch by mistake, since it's a small change I will not revert it.

entangled90:
I tested a week ago with the benchmarks from main, so I'm not sure it's up to date.

entangled90:
From the helm charts, the service is named "${orchestration.fullName}-gateway" and it's still defaulting to `zeebe` I think

```
{{- define "orchestration.fullname" -}}
    {{- /* NOTE: The value is set to "zeebe" for backward compatibility between 8.7 and 8.8. */ -}}
    {{- include "camundaPlatform.componentFullname" (dict
        "componentName" "zeebe"
        "componentValues" .Values.orchestration
        "context" $
    ) -}}
{{- end -}}
```

entangled90:
You're correct, there is another `benchmark` specific service that has the label I added there, while the `camunda-platform` helm chart uses the `component=zeebe-broker` for the services as well. 
@ChrisKujawa is this correct? sound a bit off to me, but I don't have enough context.

This is the service description from a pure camunda-platform helm chart deployed by @deepthidevaki :
```
Name:              camunda-zeebe-gateway
Namespace:         dd-test-helm-labels
Labels:            app=camunda-platform
                   app.kubernetes.io/component=zeebe-broker
                   app.kubernetes.io/instance=camunda
                   app.kubernetes.io/managed-by=Helm
                   app.kubernetes.io/name=camunda-platform
                   app.kubernetes.io/part-of=camunda-platform
                   app.kubernetes.io/version=8.8.0-rc2
                   helm.sh/chart=camunda-platform-13.0.0-alpha8
Annotations:       cloud.google.com/neg: {"ingress":true}
                   meta.helm.sh/release-name: camunda
                   meta.helm.sh/release-namespace: dd-test-helm-labels
Selector:          app.kubernetes.io/component=zeebe-broker,app.kubernetes.io/instance=camunda,app.kubernetes.io/managed-by=Helm,app.kubernetes.io/name=camunda-platform,app.kubernetes.io/part-of=camunda-platform,app=camunda-platform
Type:              ClusterIP
IP Family Policy:  SingleStack
IP Families:       IPv4
IP:                10.4.7.154
IPs:               10.4.7.154
Port:              server  9600/TCP
TargetPort:        9600/TCP
Endpoints:         10.0.47.72:9600,10.0.50.55:9600,10.0.62.10:9600
Port:              http  8080/TCP
TargetPort:        8080/TCP
Endpoints:         10.0.47.72:8080,10.0.50.55:8080,10.0.62.10:8080
Port:              grpc  26500/TCP
TargetPort:        26500/TCP
Endpoints:         10.0.47.72:26500,10.0.50.55:26500,10.0.62.10:26500
Session Affinity:  None
Events:            <none>
```

entangled90:
What's the label we should use then? 
We can try with `zeebe-gateway` for Sass and  `app.kubernetes.io/instance=camunda` if we are in SM?

PR #687, opened 2025-09-18
feat: do not wait partition scaling to complete

entangled90:
When scaling partition, zbchaos doesn't wait anymore for the change to be completed. This opens the possibility to inject chaos when the scaling is happening.

The other cluster commands are not modified as it may break some existing experiment


deepthidevaki (approved):
LGTM. Does all experiments that use this command, has an explicit wait after this step?

entangled90:
At the moment just one E2E has this command and it has a wait :+1:

PR #826, opened 2026-05-06
fix: rename worker env vars to ZEEBE_* form

pranjalg13:
## Context

The previous worker.yaml config still fails on 8.7 because the property names there use a different format than 8.8+. I verified this by deploying into an 8.7 namespace and confirmed that switching to the env var names below resolves it.

## Change

Rename the affected env vars in `worker.yaml` so the same manifest works across 8.7 and 8.8+:

- `CAMUNDA_CLIENT_GRPC_ADDRESS` → `ZEEBE_GRPC_ADDRESS`
- `CAMUNDA_CLIENT_REST_ADDRESS` → `ZEEBE_REST_ADDRESS` (not used in 8.7, but kept consistent)
- `CAMUNDA_CLIENT_AUDIENCE` → `ZEEBE_TOKEN_AUDIENCE`

For 8.7, it uses the following property to resolve `_ZEEBE_` is added as a extra param in between:
- 'CAMUNDA_CLIENT_ZEEBE_GRPC_ADDRESS`

Test assertions updated accordingly.

### DIFF
--- a/go-chaos/internal/deployment_test.go
+++ b/go-chaos/internal/deployment_test.go
@@ -355,8 +355,8 @@ func Test_ShouldRenderSpringBootEnvVarsForSelfManaged(t *testing.T) {
 
 	envs := deploymentList.Items[0].Spec.Template.Spec.Containers[0].Env
 	assert.Equal(t, "worker", envValue(envs, "SPRING_PROFILES_ACTIVE"))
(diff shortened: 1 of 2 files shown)


rodrigo-lourenco-lopes (approved):
Looks good to me, thanks for the fix!

rodrigo-lourenco-lopes:
I can test the QA in 8.7 after this is released.

PR #924, opened 2026-08-19
Replace layer 1/2 SVGs with numbered mermaid sequence diagrams

ChrisKujawa:
Jon flagged that the `07-job-push-end-to-end.svg` and `08-job-push-poll-mechanism.svg` figures in the [worker downtime / throughput recovery post](https://camunda.github.io/zeebe-chaos/2026/08/06/worker-downtime-throughput-recovery#layer-1-push-hands-out-jobs-that-never-enter-the-backlog) were hard to follow, no clear order for where to look first.

## What changed

Both diagrams are fundamentally message-ordering problems between a small set of actors (broker/gateway/worker for layer 1, push/poll/semaphore for layer 2), so they're a better fit for mermaid's autonumbered `sequenceDiagram` than hand-authored SVG. Autonumbering gives step ordering for free, and it matches the concrete-trace `sequenceDiagram` already further down this same post.

**Layer 1** now reads as a single numbered flow: jobs created during the outage land in the backlog, the worker reconnects, a new job gets pushed straight through with no backlog check, poll competes for the same capacity, and the two failure/recovery paths (fast gateway `YIELD` vs. a rare slow client-side drop) are each called out in a highlighted block.

**Layer 2** now makes one point at a time instead of everything at once: push and poll are independent paths with no coordination between them, both converge on one shared non-fair semaphore, and push filling that semaphore to capacity while poll's earlier request keeps losing is shown as explicit steps rather than asserted in prose. Blind budget and sequential dispatch are named and shown in the order they actually stack, and the closing note is careful not to overstate the push/poll difference: push blocks on the same semaphore too and its own deadline is at risk too, the real difference is that a stuck push never delays another push, while poll's batch has job 2..N queued behind job 1 on one shared deadline. Leaked budget is left out of this figure since it's rare and already covered by prose plus the existing concrete-trace diagram.

Removes the now-orphaned `07-job-push-end-to-end.svg` and `08-job-push-poll-mechanism.svg`.

## Verification

Iterated on both diagrams with the reporter across several rounds via a live mermaid preview (rendered, read back, and adjusted each time) before committing them here, per the repo's diagram-authoring guidance. Not yet checked against a live `docusaurus start` render in this environment.

## Open question

The existing concrete-trace `sequenceDiagram` further down the Layer 2 section (the one walking through blind budget → sequential dispatch → leaked budget → the rejected-completion outcome) now has some overlap with the new Layer 2 diagram in its opening two steps. It's not fully redundant, it's the only place showing the actual duplicate-completion outcome, but trimming its now-repeated opening (the blind-budget request and the first `execute(job 1)`/`execute(job 2)` pair) is worth considering as a follow-up in this same PR.


chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
```diff
@@ -177,7 +177,41 @@ The broker's own counters make this concrete. The panel above plots job creation
...
+    GPo->>B: forwarded to the broker
+    B-->>GPo: batch for A-D: priority first, then oldest key
+    Note right of B: response only once this log position is processed
+    GPo-->>W: relayed back, competing for the same permits
+
+    rect rgb(162, 59, 46, 0.12)
+    Note over GPu,B: fast recovery: push blocked at the gateway
+    GPu--xB: not ready, fails now, no queue
+    B->>B: YIELD, back to ACTIVATABLE in roughly one log write
+    end
```
multani: I'm not sure to follow what this means. My understanding:

1. This block describes a different scenario? (it seems to be a "the gateway blocks pushes" failure scenario, but its explained later in the other "layer 2" section) ; it's not 100% clear if it's a follow-up of the previous sequences, or describing something else.
2. This is the same as the sequence `#1`, right? If yes, maybe we can merge these 2 together and color it as "red / failure scenario"?

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
```diff
@@ -177,7 +177,41 @@ The broker's own counters make this concrete. The panel above plots job creation
...
+
+    rect rgb(162, 59, 46, 0.12)
+    Note over W,B: slow recovery: poll job loses the permit race,<br/>rare and not seen in this run
+    W--xW: semaphore acquire times out, RejectedExecutionException
+    W->>B: broker never told, job runs down its deadline,<br/>then churns back into the backlog
+    end
```
multani: Same as above: I'm not sure how to read it in the complete sequence diagram.

It seems both this one and the previous ones are describing failure scenarios, for specific scenarios using the beginning of the sequence diagram. If that's the case, I think it could be a bit clearer if:

1. The "failure" diagrams are separated into dedicated diagrams
2. They only show the specific scenario they represent.

I think it would even be better for this specific scenario with the timeout, as it might be possible to show when the timer starts and ends, and how it affects / is affected by the push scenario.

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
```diff
@@ -194,7 +228,39 @@ We expected to find that push simply outraces polling for the worker's capacity.
...
+    Pu->>S: tryAcquire, job 2
+    S->>Pu: permit granted
+    Pu->>S: tryAcquire, job 3
+    S->>Pu: permit granted
+    Note over S: capacity reached, every permit now held by push
+    Note over Po,S: non-fair capacity: barging allowed, no ordering<br/>guaranteed, poll's earlier request still isn't granted
+
+    Note over Po: batch arrives ACTIVATED, sharing one deadline
+    Po->>S: execute(job 1)
+    Po->>S: execute(job 2)
```
multani: Does the diagram miss a `S->>Po: tryAcquire returns X jobs`?

The diagram shows an initial `tryAcquire` but doesn't show what it returns.

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
```diff
@@ -194,7 +228,39 @@ We expected to find that push simply outraces polling for the worker's capacity.
...
+    Note right of S: blocks per job, burning the shared deadline
```
multani, line 268: Is there a way to visualize this "shared deadline"?

multani (commented):
I think the new diagrams are easier to follow (by being top-down + having the sequence numbers)

I tried to put comments on what was not super clear ; it's hard to make better suggestion since I'm not sure I understood the original diagrams correctly :)

I think the already existing sequence diagram is quite clear (at least, I have the feeling I understand it better) ; comparatively, it seems it only shows a very specific scenario (the timeout on poll) (except for the rare branch in the middle) which may explain why it seems easier to follow.

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
ChrisKujawa: 1. Layer 2 is about the client defects - here we describing the failure handling on the gateway - so when no job can be pushed out by the gateway (but already has been received by the gateway)
2. It is not the same as `#1` as here the broker can't push to the gateway - here it was pushed to the gateway but came back with a yield


Maybe we can add a sequence from Broker to GW showing a push and then a response back (including the YIELD, saying no forward possible or something)

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
ChrisKujawa: Probably makes sense to separate this a little 👍🏼

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
ChrisKujawa: Hm I think I had a diagram for this before 🤔 let me check

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
ChrisKujawa: Ok this is misleading 👍🏼 the idea was to show that it can for ever wait there but i agree i should use the similar method names - for polling it should be execute - which does the tryAcquire then

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
ChrisKujawa: @multani wdyt about this 

Image

multani: Looks much clearer, except for the `acquires here`:
* can you also add the same for the other jobs (I suppose it should be compared to the red line in the middle)
* also mention who acquires and what ; I think it would clarify that specific point

ChrisKujawa: Actually it is about another worker - tried to clarify this

multani: Ah, good that I asked :facepalm: 

This looks good to me :+1:

chaos-days/blog/2026-08-06-worker-downtime-throughput-recovery/index.md
```diff
@@ -177,24 +177,99 @@ The broker's own counters make this concrete. The panel above plots job creation
...
+    participant GPu as Gateway<br/>Push forwarding
+    participant GPo as Gateway<br/>Poll forwarding
+    participant W as Worker<br/>capacity permits
+
+    B->>GPu: Job X created, pushed
+
+    rect rgb(162, 59, 46, 0.12)
+    Note over GPu,B: push blocked at the gateway
+    GPu--xB: YIELD back, no queue
+    B->>B: YIELD puts job into ACTIVATABLE pool
```
multani, line 227:
```suggestion
    B->>B: YIELD puts job into ACTIVATABLE pool
    Note over GPu,B: From now on, this job is only available via polling
```

multani (approved):
With the new PNG you shared, this should be good.

I added a small note to make the new sequence diagram self-contained.

